PR #223, opened 2021-12-15
Update rainbow.py

saikrishna-1996:
fix `device` in rainbow initializations


dapatil211:
Not necessary, it's done in the super constructor

PR #258, opened 2022-03-12
Rnn support: DRQN agent + recurrent buffer

hnekoeiq:


hive/replays/recurrent_replay.py
```diff
@@ -0,0 +1,243 @@
...
+            array: array to access from
+            indices: starts of ranges to access from
+            num_to_access: how many consecutive elements to access
+        """
+        full_indices = np.indices((indices.shape[0], num_to_access))[1]
+        full_indices = (full_indices + np.expand_dims(indices, axis=1)) % (
+            self.size() + self._max_seq_len + self._n_step - 1
+        )
+        elements = array[full_indices]
+        elements = elements.reshape(indices.shape[0], -1, *elements.shape[2:])
```
TongTongX, line 133: We want to ensure the dimension of observation is (batch_size, seq_length, C, H, W). C=1 according to https://github.com/chandar-lab/RLHive/blame/main/hive/envs/atari/atari.py#L62.

hive/replays/recurrent_replay.py
```diff
@@ -200,7 +200,7 @@ def sample(self, batch_size):
             trajectory_lengths = (
                 np.argmax(terminals.astype(bool), axis=1) + 1
             ) * is_terminal + self._n_step * (1 - is_terminal)
-            is_terminal = terminals[:,1:self._n_step-1]
```
karthiks1701: Because of the sequence length, is_terminal has to be of shape (B*seq_length). Hence this particular change after calculating trajectory lengths.

hive/replays/recurrent_replay.py
```diff
@@ -226,18 +226,19 @@ def sample(self, batch_size):
...
                 if self._n_step > 1:
-                    idx = np.arange(rewards.shape[1]-self._n_step+1)[:,None] + np.arange(self._n_step)
-                    disc_rewards = np.einsum("ijk,k->ij",rewards[:,idx],self._discount)
+                    idx = np.arange(rewards.shape[1] - self._n_step + 1)[
+                        :, None
+                    ] + np.arange(self._n_step)
+                    disc_rewards = np.einsum(
+                        "ijk,k->ij", rewards[:, idx], self._discount
+                    )
                     rewards = disc_rewards
```
karthiks1701, line 250: [Creating a vectorized sliding window to calculate discounted returns for every element in the sequence. ](https://stackoverflow.com/questions/39232790/numpy-vectorization-of-sliding-window-operation)

hive/agents/qnets/qnet_heads.py
```diff
@@ -39,6 +39,43 @@ def forward(self, x):
         return self.output_layer(x)
 
 
+class DRQNNetwork(nn.Module):
+    """Implements the standard DQN value computation. Transforms output from
+    :obj:`base_network` with output dimension :obj:`hidden_dim` to dimension
+    :obj:`out_dim`, which should be equal to the number of actions.
+    """
```
dapatil211: Update docstring

hive/agents/qnets/qnet_heads.py
```diff
@@ -39,6 +39,43 @@ def forward(self, x):
...
+        self,
+        base_network: nn.Module,
+        hidden_dim: int,
+        out_dim: int,
+        linear_fn: nn.Module = None,
+    ):
+        """
+        Args:
+            base_network (torch.nn.Module): Backbone network that computes the
+                representations that are used to compute action values.
```
dapatil211: Update docstring

hive/agents/drqn.py
```diff
@@ -0,0 +1,308 @@
...
+    Schedule,
+    SwitchSchedule,
+)
+from hive.agents.dqn import DQNAgent
+from hive.utils.utils import LossFn, OptimizerFn, create_folder, seeder
+
+
+class DRQNAgent(DQNAgent):
+    """An agent implementing the DQN algorithm. Uses an epsilon greedy
+    exploration policy
```
dapatil211: Docstring

hive/agents/qnets/rnn.py
```diff
@@ -0,0 +1,149 @@
...
+        self.lstm = nn.LSTM(
+            np.prod(conv_output_size),
+            lstm_hidden_size,
+            num_lstm_layers,
+            batch_first=True,
+        )
+
+        if mlp_layers is not None:
+            # MLP Layers
+            # conv_output_size = calculate_output_dim(self.conv, in_dim)
```
dapatil211: delete

hive/agents/qnets/utils.py
```diff
@@ -20,9 +20,12 @@ def calculate_output_dim(net, input_shape):
     """
     if isinstance(input_shape, int):
         input_shape = (input_shape,)
-    placeholder = torch.zeros((0,) + tuple(input_shape))
+    placeholder = torch.zeros((1,) + tuple(input_shape)).to(device)
```
dapatil211: Device shouldn't need to be passed in right? Like at this point it should all be on the cpu? I am fine with the change, just not sure why it's necessary

hive/agents/qnets/utils.py
```diff
@@ -20,9 +20,12 @@ def calculate_output_dim(net, input_shape):
...
     output = net(placeholder)
-    return output.size()[1:]
+    if isinstance(output, tuple):
```
dapatil211: If it's a tuple, you should return the size of each output. It doesn't make sense to only do the first output. That's just specific to your current use case

hive/replays/recurrent_replay.py
```diff
@@ -0,0 +1,262 @@
...
+                        :, None
+                    ] + np.arange(self._n_step)
+                    disc_rewards = np.einsum(
+                        "ijk,k->ij", rewards[:, idx], self._discount
+                    )
+                    rewards = disc_rewards
+
+                # Mask out rewards past trajectory length
+                # mask = np.expand_dims(trajectory_lengths, 1) > np.arange(self._n_step)
+                # rewards = np.sum(rewards * mask, axis=1)
```
dapatil211: delete

hive/replays/recurrent_replay.py
```diff
@@ -0,0 +1,262 @@
...
+                    rewards = np.expand_dims(rewards, 1)
+
+                if self._n_step == 1:
+                    rewards = rewards * np.expand_dims(self._discount, axis=0)
+
+                if self._n_step > 1:
+                    idx = np.arange(rewards.shape[1] - self._n_step + 1)[
+                        :, None
+                    ] + np.arange(self._n_step)
+                    disc_rewards = np.einsum(
```
dapatil211: Could you give a comment as to what this is doing? einsum is notoriously bad in terms of being interpreted.

dapatil211 (changes requested):
I went through and gave a quick review. I didn't thoroughly check for correctness yet, ideally would want to see experiments that show it's working first before doing that. Also please write updated docstrings.

hive/replays/recurrent_replay.py
karthiks1701: Sure will do!

hive/agents/qnets/utils.py
hnekoeiq: @dapatil211 The problem is here we either need to create a dummy `hidden_state` on CPU and pass it to the network or add an if condition in the forward function of ConvRNNNetwork to take care of `hidden_state = None`. We did the latter but there might be another case other than `calculate_output_dim` where `hidden_state` is None and in that case, our `hidden_state` inside the forward function should be on the _device. (We had to change the empty array because lstm was not able to process it)

dapatil211: Well don't you only need to do that because you are doing `.to(device)` here when you don't really need to?: https://github.com/chandar-lab/RLHive/pull/258/files/4874d7a579fbcaf3c26dce7d3fecc01c5c57fac6#diff-324043212c24b1c8e9c8c139a4defa49947886ee8fc9f4f9ab8138d4347f4860R139-R141

hive/agents/qnets/rnn.py
hnekoeiq: Done

hive/agents/qnets/utils.py
hnekoeiq: fixed

hnekoeiq:
Results: 
[DRQN vs DQN on Pong, DoubleDunk, Frostbite](https://wandb.ai/chandar-rl/Hive-rnn/reports/Hive-RNN-support--VmlldzoxNzAxNDgz?accessToken=vmfly5b5wfo3iyttea3v3pveqpyzuh72j0ag1gkplz3ayxs815mv7e2o4rut5j7c)

[DRQN vs DQN on Frostbite and Beam Rider](https://wandb.ai/somz/RNN/reports/Beamrider--VmlldzoxNzEyMDQ3)

hive/agents/drqn.py
```diff
@@ -0,0 +1,307 @@
...
+                during learning.
+            device: Device on which all computations should be run.
+            logger (ScheduledLogger): Logger used to log agent's metrics.
+            log_frequency (int): How often to log the agent's metrics.
+        """
+        super().__init__(
+            representation_net=representation_net,
+            obs_dim=obs_dim,
+            act_dim=act_dim,
+            id=0,
```
karthiks1701: I think we missed `id=id`. to pass the id to the base DQN class, because of this in marlgrid all the agents have the same id and that is causing problems.

hive/agents/qnets/utils.py
hnekoeiq: @dapatil211 This one needs a bit more discussion. Since `hidden_state` (the second item in the tuple) is a tuple itself, should we check the items inside that and return their size as well?

dapatil211: Probably should be recursive? Keep going until you hit scalars or tensors

hive/agents/drqn.py
```diff
@@ -0,0 +1,309 @@
...
+            target_net_update_fraction=target_net_update_fraction,
+            target_net_update_schedule=target_net_update_schedule,
+            epsilon_schedule=epsilon_schedule,
+            test_epsilon=test_epsilon,
+            min_replay_history=min_replay_history,
+            batch_size=batch_size,
+            device=device,
+            logger=logger,
+            log_frequency=log_frequency,
+        )
```
dapatil211: If your init function isn't doing anything other than calling the super constructor, then you can probably just delete it

hive/agents/drqn.py
```diff
@@ -0,0 +1,309 @@
...
+    def create_q_networks(self, representation_net):
+        """Creates the Q-network and target Q-network.
+
+        Args:
+            representation_net: A network that outputs the representations that will
+                be used to compute Q-values (e.g. everything except the final layer
+                of the DQN).
+        """
+        network = representation_net(self._obs_dim)
+        network_output_dim = np.prod(calculate_output_dim(network, self._obs_dim))
```
dapatil211: If you do change calculate_output_dim to return all the shapes, then you can/should add a check to make sure that the representation_net is giving the appropriate number/shape of outputs.

hive/agents/qnets/qnet_heads.py
```diff
@@ -39,6 +39,43 @@ def forward(self, x):
         return self.output_layer(x)
 
 
+class DRQNNetwork(nn.Module):
+    """Implements the standard DRQN value computation. Transforms output from
+    :obj:`base_network` with output dimension :obj:`hidden_dim` to dimension
```
dapatil211: base_network has two outputs. That should be mentioned, as well as which output you take

hive/configs/atari/drqn.yml
```diff
@@ -0,0 +1,82 @@
+run_name: &run_name 'atari-dqn'
```
dapatil211: run name should be atari-drqn

hive/replays/recurrent_replay.py
```diff
@@ -0,0 +1,261 @@
...
+        observation_shape=(),
+        observation_dtype=np.uint8,
+        action_shape=(),
+        action_dtype=np.int8,
+        reward_shape=(),
+        reward_dtype=np.float32,
+        extra_storage_types=None,
+        num_players_sharing_buffer: int = None,
+    ):
+        """Constructor for CircularReplayBuffer.
```
dapatil211: Name

hive/replays/recurrent_replay.py
```diff
@@ -0,0 +1,261 @@
...
+                    indices - self._max_seq_len + 1,
+                    num_to_access=self._max_seq_len + self._n_step - 1,
+                )
+                if self._max_seq_len + self._n_step - 1 == 1:
+                    rewards = np.expand_dims(rewards, 1)
+
+                if self._n_step == 1:
+                    rewards = rewards * np.expand_dims(self._discount, axis=0)
+
+                if self._n_step > 1:
```
dapatil211: should be an else (or elif)

hive/replays/recurrent_replay.py
```diff
@@ -0,0 +1,261 @@
...
+                    idx = np.arange(rewards.shape[1] - self._n_step + 1)[
+                        :, None
+                    ] + np.arange(self._n_step)
+
+                    # Creating a vectorized sliding window to calculate discounted returns for every element in the sequence
+                    disc_rewards = np.einsum(
+                        "ijk,k->ij", rewards[:, idx], self._discount
+                    )
```
dapatil211: So this section is still a bit confusing. I had to work it out separately to make sense of what was happening. Could you write it maybe something like this?:
```
idx = np.arange(rewards.shape[1] - self._n_step + 1)[:, None] # (S-N+1) x 1
idx = idx + np.arange(self._n_step) # (S-N+1) x N
rewards = rewards[idx] # B x (S-N+1) x N
# equivalent to np.sum(rewards * self._discount[None, None, :], axis=2) 
rewards = np.einsum("ijk,k->ij", rewards, self._discount)
```
Above is just a suggestion, you can change it to make it look better. Maybe also combine first two lines. But idea is to make it clear what's happening.

dapatil211 (changes requested):

hive/agents/drqn.py
```diff
@@ -0,0 +1,208 @@
...
+    """
+
+    def create_q_networks(self, representation_net):
+        """Creates the Q-network and target Q-network.
+
+        Args:
+            representation_net: A network that outputs the representations that will
+                be used to compute Q-values (e.g. everything except the final layer
+                of the DRQN).
+        """
```
dapatil211, line 155: You should add a comment about the expected output of this network, ie that it outputs a output and hidden state

hive/agents/drqn.py
```diff
@@ -0,0 +1,208 @@
...
+            and self._logger.should_log(self._timescale)
+            and self._state["episode_start"]
+        ):
+            self._logger.log_scalar("train_qval", torch.max(qvals), self._timescale)
+            self._state["episode_start"] = False
+        return action
+
+    def update(self, update_info):
+        """
+        Updates the DQN agent.
```
dapatil211: Update

hive/agents/qnets/qnet_heads.py
```diff
@@ -39,6 +39,43 @@ def forward(self, x):
         return self.output_layer(x)
 
 
+class DRQNNetwork(nn.Module):
+    """Implements the standard DRQN value computation. Transforms output from
+    :obj:`base_network` with output dimension :obj:`hidden_dim` and previous hidden_state to dimension
+    :obj:`out_dim`, which should be equal to the number of actions and next hidden_state which will be used as initial hidden_state later.
```
dapatil211: Make sure docstrings/comments aren't too long

hive/agents/qnets/qnet_heads.py
dapatil211: You should say something about the expected output of this base_network

hive/agents/qnets/utils.py
```diff
@@ -20,9 +20,18 @@ def calculate_output_dim(net, input_shape):
     """
     if isinstance(input_shape, int):
         input_shape = (input_shape,)
-    placeholder = torch.zeros((0,) + tuple(input_shape))
+    placeholder = torch.zeros((1,) + tuple(input_shape))
```
dapatil211, line 23: why was this changed?

hive/replays/recurrent_replay.py
```diff
@@ -0,0 +1,262 @@
...
+                    rewards = rewards * np.expand_dims(self._discount, axis=0)
+
+                elif self._n_step > 1:
+                    idx = np.arange(rewards.shape[1] - self._n_step + 1)[
+                        :, None
+                    ] + np.arange(
+                        self._n_step
+                    )  # (S-N+1) x N
+                    rewards = rewards[:, idx]  # B x (S-N+1) x N
+                    # Creating a vectorized sliding window to calculate discounted returns for every element in the sequence
```
dapatil211: comment line length too long

dapatil211 (changes requested):
Mostly comment/docstring related changes, a couple other issues also marked. Also, could you add this to the tests already created or write new tests if needed?

hive/agents/drqn.py
```diff
@@ -0,0 +1,208 @@
...
+                batch_size=self._batch_size, device=self._device
+            )
+            target_hidden_state = self._target_qnet.base_network.init_hidden(
+                batch_size=self._batch_size, device=self._device
+            )
+            # Compute predicted Q values
+            self._optimizer.zero_grad()
+            pred_qvals, hidden_state = self._qnet(*current_state_inputs, hidden_state)
+            pred_qvals = pred_qvals.view(
+                self._batch_size, self._replay_buffer._max_seq_len, -1
```
dapatil211: Should not be accessing internal variable of buffer

hive/agents/qnets/rnn.py
```diff
@@ -0,0 +1,168 @@
...
+                conv_seq.append(torch.nn.ReLU())
+            self.conv = torch.nn.Sequential(*conv_seq)
+        else:
+            self.conv = torch.nn.Identity()
+
+        # RNN Layers
+        conv_output_size = calculate_output_dim(self.conv, in_dim)
+        if self._rnn_type == "lstm":
+            self.rnn = nn.LSTM(
+                np.prod(conv_output_size),
```
sriyash421: Changing different types of network to a sequencer class

hive/agents/drqn.py
```diff
@@ -0,0 +1,208 @@
...
+        """
+        Updates the DQN agent.
+
+        Args:
+            update_info: dictionary containing all the necessary information to
+                update the agent. Should contain a full transition, with keys for
+                "observation", "action", "reward", and "done".
+        """
+        if update_info["done"]:
+            self._state["episode_start"] = True
```
sriyash421, line 223: is episode start used anywhere except the buffer? because the buffer takes care of it and it is redundant.

hive/agents/qnets/sequence_models.py
```diff
@@ -0,0 +1,145 @@
...
+    ):
+        """
+        Args:
+            rnn_input_size (int): The number of expected features in the input x.
+            rnn_hidden_size (int):  The number of features in the hidden state h.
+            num_rnn_layers (int): Number of recurrent layers.
+            batch_first (bool): If True, then the input and output tensors are provided as (batch, seq, feature) instead of (seq, batch, feature).
+        """
+        super().__init__(
+            rnn_input_size=rnn_input_size,
```
hnekoeiq: You can remove rnn_input_size if the base function does not use it.

TongTongX: Done. SImilarly batch_first is also removed.

hive/agents/qnets/rnn.py
TongTongX: Done.

hive/agents/__init__.py
```diff
@@ -16,6 +17,7 @@
         "LegalMovesRainbowAgent": LegalMovesRainbowAgent,
         "RainbowDQNAgent": RainbowDQNAgent,
         "RandomAgent": RandomAgent,
+        "DRQNAgent": DRQNAgent,
```
dapatil211: Alphabetical order please.

hive/agents/drqn.py
```diff
@@ -0,0 +1,283 @@
...
+from hive.replays.recurrent_replay import RecurrentReplayBuffer
+from hive.utils.loggers import Logger, NullLogger
+from hive.utils.schedule import (
+    LinearSchedule,
+    PeriodicSchedule,
+    Schedule,
+    SwitchSchedule,
+)
+from hive.agents.dqn import DQNAgent
+from hive.utils.utils import LossFn, OptimizerFn, create_folder, seeder
```
dapatil211: Make sure you run import sorting (`isort`)

hive/agents/drqn.py
```diff
@@ -0,0 +1,283 @@
...
+        batch_size: int = 32,
+        device="cpu",
+        logger: Logger = None,
+        log_frequency: int = 100,
+    ):
+        """
+        Args:
+            observation_space (gym.spaces.Box): Observation space for the agent.
+            action_space (gym.spaces.Discrete): Action space for the agent.
+            representation_net (FunctionApproximator): A network that outputs the
```
dapatil211, line 67: I am assuming there are restrictions on the representation_net? eg it needs to be one of your recurrent ones? Please mention this in the documentation.

hive/agents/drqn.py
```diff
@@ -0,0 +1,283 @@
...
+                representations that will be used to compute Q-values (e.g.
+                everything except the final layer of the DRQN).
+            stack_size: Number of observations stacked to create the state fed to the
+                DRQN.
```
dapatil211: What does this mean in the context of DRQN? You aren't using this are you? If not, remove. You may need to keep a dummy or add a kwargs to make it work with the runner.

hive/agents/drqn.py
```diff
@@ -0,0 +1,283 @@
...
+from hive.agents.dqn import DQNAgent
+from hive.utils.utils import LossFn, OptimizerFn, create_folder, seeder
+
+
+class DRQNAgent(DQNAgent):
+    """An agent implementing the DRQN algorithm. Uses an epsilon greedy
+    exploration policy
+    """
+
+    def __init__(
```
dapatil211, line 35: Please go through the arguments and documentation and make sure the ones you are exposing are all actually being used by your agent.

hive/agents/drqn.py
```diff
@@ -0,0 +1,283 @@
...
+            logger=logger,
+            log_frequency=log_frequency,
+        )
+        if replay_buffer is None:
+            replay_buffer = RecurrentReplayBuffer
+        self._replay_buffer = replay_buffer(
+            max_seq_len=max_seq_len,
+            observation_shape=self._observation_space.shape,
+            observation_dtype=self._observation_space.dtype,
+        )
```
dapatil211: you should also probably add action_shape and action_dtype. This will need to be done in DQN as well (in a separate PR)

hive/agents/drqn.py
dapatil211: This is there for stuff in act(). There's probably a better way to do it. It doesn't really make sense for the agent to do it. It might make sense to add it as part of the observation, but need to think about this a bit.

hive/agents/drqn.py
```diff
@@ -0,0 +1,283 @@
...
+        Args:
+            update_info: dictionary containing all the necessary information to
+                update the agent. Should contain a full transition, with keys for
+                "observation", "action", "reward", and "done".
+        """
+        if update_info["done"]:
+            self._state["episode_start"] = True
+            self._hidden_state = self._qnet.base_network.init_hidden(
+                batch_size=1, device=self._device
+            )
```
dapatil211: This should probably be moved to act()

hive/agents/drqn.py
```diff
@@ -0,0 +1,283 @@
...
+
+            hidden_state = self._qnet.base_network.init_hidden(
+                batch_size=self._batch_size, device=self._device
+            )
+            target_hidden_state = self._target_qnet.base_network.init_hidden(
+                batch_size=self._batch_size, device=self._device
+            )
+            # Compute predicted Q values
+            self._optimizer.zero_grad()
+            pred_qvals, hidden_state = self._qnet(*current_state_inputs, hidden_state)
```
dapatil211: Are these hidden_states actually used elsewhere in the code? If not, might be worth changing to `pred_qvals, _ = self.qnet()`. Same for target

hive/agents/drqn.py
```diff
@@ -0,0 +1,283 @@
...
+            )
+            actions = batch["action"].long()
+            pred_qvals = torch.gather(pred_qvals, -1, actions.unsqueeze(-1)).squeeze(-1)
+
+            # Compute 1-step Q targets
+            next_qvals, target_hidden_state = self._target_qnet(
+                *next_state_inputs, target_hidden_state
+            )
+            next_qvals = next_qvals.view(
+                self._batch_size, self._replay_buffer._max_seq_len, -1
```
dapatil211: Should not be accessing internal variable of replay_buffer

hive/agents/qnets/qnet_heads.py
```diff
@@ -39,6 +39,44 @@ def forward(self, x):
         return self.output_layer(x)
 
 
+class DRQNNetwork(nn.Module):
+    """Implements the standard DRQN value computation. Transforms first output from
+    :obj:`base_network` with output dimension :obj:`hidden_dim` to dimension
+    :obj:`out_dim`, which should be equal to the number of actions.
+    The second output is the next hidden_state which will be used as initial hidden_state later.
```
dapatil211: I think this needs to be clarified a bit. Explicitly mention that this module returns two things, and that base_network should also return two things. Also that the second thing that base_network returns is the hidden state that will be used as the initial hidden state when computing the next action in the trajectory. I don't think it makes sense to call it the initial hidden_state.

hive/agents/qnets/utils.py
dapatil211: Any answer to this question?

hive/agents/qnets/qnet_heads.py
dapatil211: Please update documentation according to previous comment. Specifically that base_network returns two things, and what those two things are.

hive/agents/qnets/qnet_heads.py
```diff
@@ -39,6 +39,44 @@ def forward(self, x):
...
+        super().__init__()
+        self.base_network = base_network
+        self._linear_fn = linear_fn if linear_fn is not None else nn.Linear
+        self.output_layer = self._linear_fn(hidden_dim, out_dim)
+
+    def forward(self, x, hidden_state=None):
+        x, hidden_state = self.base_network(x, hidden_state)
+
+        x = x.flatten(start_dim=1)
+        return self.output_layer(x), hidden_state
```
dapatil211, line 81: Is the only place that this returned hidden_state used in the act() function of the agent?

hive/agents/qnets/sequence_models.py
```diff
@@ -0,0 +1,135 @@
+import torch
+from torch import nn
+
+from hive.utils.registry import registry, Registrable
+
+
+class SequenceModule(nn.Module, Registrable):
```
dapatil211: I don't really see what the point of this class is. It's not really providing any additional logic. I think instead of this, you can just do:
```
registry.register_all(FunctionApproximator,
{
'LSTM':FunctionApproximator(torch.nn.LSTM),
'GRU':FunctionApproximator(torch.nn.GRU),
}
```

In the convrnnnetwork class, you can just have it take in a FunctionApproximator for the seqence_fn and create it directly. If you really want to differentiate SequenceModels as a type, you should just have it subclass Registrable and provide a type name, similar to what FunctionApproximator does. It shouldn't subclass nn.Module.

hive/agents/qnets/rnn.py
```diff
@@ -0,0 +1,136 @@
...
+    convolution portion of the architecture is simply an
+    :py:class:`torch.nn.Identity` module. If :obj:`mlp_layers` is
+    :const:`None`, the mlp portion of the architecture is an
+    :py:class:`torch.nn.Identity` module.
+    """
+
+    def __init__(
+        self,
+        in_dim,
+        sequence_fn: SequenceModule,
```
dapatil211: See my comment in the sequence_models.py file about this.

hive/replays/recurrent_replay.py
```diff
@@ -0,0 +1,262 @@
...
+        n_step: int = 1,
+        gamma: float = 0.99,
+        observation_shape=(),
+        observation_dtype=np.uint8,
+        action_shape=(),
+        action_dtype=np.int8,
+        reward_shape=(),
+        reward_dtype=np.float32,
+        extra_storage_types=None,
+        num_players_sharing_buffer: int = None,
```
dapatil211, line 26: Why is the spacing inconsistent?

hive/replays/recurrent_replay.py
```diff
@@ -0,0 +1,262 @@
...
+    ):
+        """Constructor for RecurrentReplayBuffer.
+
+        Args:
+            capacity (int): Total number of observations that can be stored in the
+                buffer. Note, this is not the same as the number of transitions that
+                can be stored in the buffer.
+            max_seq_len (int): The number of consecutive transitions in a sequence.
```
dapatil211: Please clarify the description of this parameter.

hive/replays/recurrent_replay.py
```diff
@@ -0,0 +1,262 @@
...
+            observation_dtype=observation_dtype,
+            action_shape=action_shape,
+            action_dtype=action_dtype,
+            reward_shape=reward_shape,
+            reward_dtype=reward_dtype,
+            extra_storage_types=extra_storage_types,
+            num_players_sharing_buffer=num_players_sharing_buffer,
+        )
+        self._max_seq_len = max_seq_len
+
```
dapatil211, line 71: Would this class be equivalent to just creating CircularReplayBuffer(stack_size=max_seq_len)? If so, why not just do that? If not, what/where is the change in logic?

hive/replays/recurrent_replay.py
dapatil211: Please make sure commented lines and lines part of docstrings are still under 88 characters.

dapatil211 (changes requested):

hive/agents/qnets/utils.py
TongTongX: The first dimension is the batch size. Technically it doesn't affect the output dim since extract_shape truncates the output. I remember we changed this dimension to 1 from 0 because the output hidden state had wrong dimensions. @hnekoeiq was that the reason?

hive/agents/drqn.py
TongTongX: We probably want to fix it in a separate PR as this is what DQN does too.

hive/agents/qnets/qnet_heads.py
TongTongX: Right. `forward()` is also called in `update()` of the agent but the output hidden_state is not used.

hive/replays/recurrent_replay.py
TongTongX: If we specify the datatype black will add whitespaces around the equal sign, otherwise it will remove those whitespaces.

hive/replays/recurrent_replay.py
TongTongX: There are differences in `sample()`.

hive/agents/qnets/sequence_models.py
TongTongX: Although `SequenceModule` is simply a wrapper for LSTM or GRU for now, we made it a separate class mainly because we can implement our own recurrent module later on.

hive/agents/qnets/utils.py
hnekoeiq: If I remember correctly, I had to do this because of RNN module.

hive/runners/single_agent_loop.py
```diff
@@ -153,7 +153,6 @@ def set_up_experiment(config):
     agent = agent_fn(
         observation_space=env_spec.observation_space[0],
         action_space=env_spec.action_space[0],
-        stack_size=config.get("stack_size", 1),
```
dapatil211: This shouldn't be removed. It will break other agents.

dapatil211 (changes requested):

hive/agents/drqn.py
TongTongX: We created an [issue](https://github.com/chandar-lab/RLHive/issues/297) for fixing it in both DQN and DRQN.

hive/agents/qnets/qnet_heads.py
TongTongX: done

hive/agents/qnets/qnet_heads.py
TongTongX: done

hive/agents/drqn.py
TongTongX: done

hive/replays/recurrent_replay.py
TongTongX: done

hive/agents/drqn.py
TongTongX: done

hive/agents/qnets/qnet_heads.py
TongTongX: done

hive/configs/atari/drqn.yml
TongTongX: done

hive/agents/drqn.py
TongTongX: done

hive/agents/drqn.py
TongTongX: done

hive/agents/qnets/qnet_heads.py
TongTongX: done

hive/replays/recurrent_replay.py
TongTongX: done

hive/agents/drqn.py
TongTongX: done

hive/agents/__init__.py
TongTongX: done

hive/agents/drqn.py
TongTongX: done

hive/agents/drqn.py
TongTongX: done

hive/agents/drqn.py
TongTongX: done

hive/agents/drqn.py
TongTongX: done

hive/agents/drqn.py
TongTongX: done

hive/agents/drqn.py
TongTongX: done

hive/agents/drqn.py
TongTongX: done

hive/agents/qnets/qnet_heads.py
TongTongX: done

hive/agents/qnets/sequence_models.py
TongTongX: done

hive/agents/qnets/rnn.py
TongTongX: done

hive/replays/recurrent_replay.py
TongTongX: done

hive/runners/single_agent_loop.py
TongTongX: done

hive/agents/drqn.py
TongTongX: This is related to the `episode_start` thing above. The hidden state of RNN needs to be reinitialized at the beginning of each episode. We can fix it in the same separate PR.

hive/agents/drqn.py
dapatil211: I think this part is fine. I don't think it needs fixing.

hive/agents/drqn.py
dapatil211: I don't think this is the same thing. episode_start can only be set in update because that's when the agent knows that the episode ended. My comment is about how the resetting of the hidden state should be done in act()

hive/agents/qnets/sequence_models.py
```diff
@@ -0,0 +1,136 @@
...
+        rnn_hidden_size=128,
+        num_rnn_layers=1,
+        batch_first=True,
+    ):
+        """
+        Args:
+            rnn_input_size (int): The number of expected features in the input x.
+            rnn_hidden_size (int):  The number of features in the hidden state h.
+            num_rnn_layers (int): Number of recurrent layers.
+            batch_first (bool): If True, then the input and output tensors are provided as (batch, seq, feature) instead of (seq, batch, feature).
```
dapatil211: line length

hive/agents/qnets/sequence_models.py
```diff
@@ -0,0 +1,136 @@
...
```
dapatil211: line length

hive/agents/qnets/sequence_models.py
```diff
@@ -0,0 +1,136 @@
...
+            num_rnn_layers=num_rnn_layers,
+        )
+        self.core = nn.LSTM(
+            input_size=rnn_input_size,
+            hidden_size=self._rnn_hidden_size,
+            num_layers=self._num_rnn_layers,
+            batch_first=batch_first,
+        )
+
+    def init_hidden(self, batch_size, device="cpu"):
```
dapatil211: can this device not just be passed once in the initializer?

dapatil211 (changes requested):

hive/agents/qnets/sequence_models.py
TongTongX: Fixed.

hive/agents/qnets/sequence_models.py
TongTongX: Fixed.

hive/agents/qnets/sequence_models.py
TongTongX: Fixed.

hive/agents/drqn.py
```diff
@@ -0,0 +1,281 @@
...
+        if replay_buffer is None:
+            replay_buffer = RecurrentReplayBuffer
+        self._replay_buffer = replay_buffer(
+            max_seq_len=max_seq_len,
+            observation_shape=self._observation_space.shape,
+            observation_dtype=self._observation_space.dtype,
+            action_shape=self._action_space.shape,
+            action_dtype=self._action_space.dtype,
+        )
+        self._max_seq_len = max_seq_len
```
dapatil211: This can be moved above the super constructor call:
```
        if replay_buffer is None:
            replay_buffer = RecurrentReplayBuffer
        replay_buffer = partial(replay_buffer, max_seq_len=max_seq_len)
```

dapatil211 (changes requested):

hive/agents/drqn.py
```diff
@@ -0,0 +1,277 @@
...
+                to optimize and returns the optimizer. If None, defaults to
+                :py:class:`~torch.optim.Adam`.
+            loss_fn (LossFn): Loss function used by the agent. If None, defaults to
+                :py:class:`~torch.nn.SmoothL1Loss`.
+            init_fn (InitializationFn): Initializes the weights of qnet using
+                create_init_weights_fn.
+            replay_buffer (BaseReplayBuffer): The replay buffer that the agent will
+                push observations to and sample from during learning. If None,
+                defaults to
+                :py:class:`~hive.replays.circular_replay.CircularReplayBuffer`.
```
dapatil211: This is incorrect. Please go through all the documentation and make sure it is correct.

hive/agents/drqn.py
dapatil211: Can you more explicitly mention the restrictions on representation_net? For example, which methods it should have or that it should follow the structure of ConvRNNNetwork or something?

hive/agents/drqn.py
```diff
@@ -0,0 +1,277 @@
...
+                next_state_inputs,
+                batch,
+            ) = self.preprocess_update_batch(batch)
+
+            hidden_state = self._qnet.base_network.init_hidden(
+                batch_size=self._batch_size, device=self._device
+            )
+            target_hidden_state = self._target_qnet.base_network.init_hidden(
+                batch_size=self._batch_size, device=self._device
+            )
```
dapatil211: It feels weird that you are accessing an internal module of the qnet. I think instead of `self._qnet.base_network.init_hidden`, it should be `self._qnet.init_hidden`

hive/agents/qnets/qnet_heads.py
```diff
@@ -39,6 +39,48 @@ def forward(self, x):
...
+                number of actions that you are computing values for.
+            linear_fn (torch.nn.Module): Function that will create the
+                :py:class:`torch.nn.Module` that will take the output of
+                :obj:`network` and produce the final action values. If
+                :obj:`None`, a :py:class:`torch.nn.Linear` layer will be used.
+        """
+        super().__init__()
+        self.base_network = base_network
+        self._linear_fn = linear_fn if linear_fn is not None else nn.Linear
+        self.output_layer = self._linear_fn(hidden_dim, out_dim)
```
dapatil211, line 75: should probably make all of these internal variables

hive/replays/recurrent_replay.py
```diff
@@ -0,0 +1,261 @@
...
+                elif self._n_step > 1:
+                    idx = np.arange(rewards.shape[1] - self._n_step + 1)[
+                        :, None
+                    ] + np.arange(
+                        self._n_step
+                    )  # (S-N+1) x N
+                    rewards = rewards[:, idx]  # B x (S-N+1) x N
+                    # Creating a vectorized sliding window to calculate
+                    # discounted returns for every element in the sequence.
+                    # equivalent to np.sum(rewards * self._discount[None, None, :], axis=2)
```
dapatil211: This line is longer than 88 characters.

dapatil211 (changes requested):

hive/agents/drqn.py
```diff
@@ -0,0 +1,277 @@
...
+            # Compute predicted Q values
+            self._optimizer.zero_grad()
+            pred_qvals, _ = self._qnet(*current_state_inputs, hidden_state)
```
sriyash421, line 254: Just to make sure, the qnet takes in a window of past observations and you take as output the last hidden state and pass it through an MLP to get the Q-values? So, it involves some redundant computation when calculating Q-values for s_t and s_t+1

hive/agents/qnets/__init__.py
```diff
@@ -3,12 +3,14 @@
...
 from hive.agents.qnets.conv import ConvNetwork
 from hive.agents.qnets.mlp import MLPNetwork
+from hive.agents.qnets.rnn import ConvRNNNetwork
 
 registry.register_all(
     FunctionApproximator,
     {
         "MLPNetwork": MLPNetwork,
         "ConvNetwork": ConvNetwork,
+        "ConvRNNNetwork": ConvRNNNetwork,
```
sriyash421, line 12: alphabetical order?

hive/agents/qnets/qnet_heads.py
sriyash421: Maybe it can have an internal function to call during act and update separately?

hive/agents/drqn.py
```diff
@@ -0,0 +1,277 @@
...
+            q_targets = batch["reward"] + self._discount_rate * next_qvals * (
+                1 - batch["done"]
+            )
+
+            loss = self._loss_fn(pred_qvals, q_targets).mean()
+
+            if self._logger.should_log(self._timescale):
+                self._logger.log_scalar("train_loss", loss, self._timescale)
+
+            loss.backward()
```
sriyash421, line 273: maybe move this line above the log step? to avoid issues with the loss if the logger function makes changes to the loss value in a call by reference setting?

hive/agents/qnets/qnet_heads.py
```diff
@@ -39,6 +39,48 @@ def forward(self, x):
...
+    transforms the first output from :obj:`base_network` with output dimension
+    :obj:`hidden_dim` to dimension :obj:`out_dim`, which should be equal to the
+    number of actions. The second output of this module is the second output from
+    :obj:`base_network`, which is the hidden state that will be used as the initial
+    hidden state when computing the next action in the trajectory.
+    """
+
+    def __init__(
+        self,
+        base_network: nn.Module,
```
sriyash421, line 54: base network should be a sequence model?

hive/agents/qnets/rnn.py
```diff
@@ -0,0 +1,138 @@
...
+                sequence_fn.keywords["rnn_hidden_size"],
+                mlp_layers,
+                noisy=noisy,
+                std_init=std_init,
+            )
+        else:
+            self.mlp = nn.Identity()
+
+    def forward(self, x, hidden_state=None):
+        if len(x.shape) == 4:
```
sriyash421: can be cleaned if we have separate call functions to act and update, as suggested above.

hive/agents/qnets/sequence_models.py
```diff
@@ -0,0 +1,145 @@
...
+class SequenceModel(nn.Module):
+    """
+    Base sequence neural network architecture.
+    """
+
+    def __init__(
+        self,
+        rnn_hidden_size=128,
+        num_rnn_layers=1,
+        device="cpu",
```
sriyash421, line 25: device is an existing param of nn.Module? do we need to redefine?

hive/replays/__init__.py
```diff
@@ -11,6 +12,7 @@
         "SimpleReplayBuffer": SimpleReplayBuffer,
         "PrioritizedReplayBuffer": PrioritizedReplayBuffer,
         "LegalMovesBuffer": LegalMovesBuffer,
+        "RecurrentReplayBuffer": RecurrentReplayBuffer,
```
sriyash421, line 14: alphabetical order please?

hive/replays/recurrent_replay.py
sriyash421: Functions and code resued from CircularReplayBuffer? Maybe just use the inherited functions? Or are there changes across all the functions?

sriyash421 (changes requested):

hive/agents/drqn.py
dapatil211: I don't think this should be an issue, since wandb and chomp both unwrap the tensor to get the raw value, but yeah shouldn't hurt

hive/replays/recurrent_replay.py
TongTongX: We are using the inherited functions defined in CircularReplayBuffer except for the ones where `max_seq_len` and `stack_size` differ.

hive/agents/drqn.py
TongTongX: done

hive/agents/drqn.py
TongTongX: done

hive/agents/drqn.py
TongTongX: done

hive/agents/drqn.py
TongTongX: done

hive/agents/qnets/qnet_heads.py
TongTongX: maybe in a separate PR? All network modules defined here have `base_network` and `output_layer`.

hive/replays/recurrent_replay.py
TongTongX: done

hive/agents/qnets/__init__.py
TongTongX: done

hive/agents/qnets/qnet_heads.py
TongTongX: It is a network module. The sequence model, conv layers, and MLP layers are components of `base_network`.

hive/agents/qnets/sequence_models.py
TongTongX: It would look messy if every time we set the device of the hidden state the same one as that of the base network. We chose the easy way that specifies the device of sequence model in the `.yml` config.

hive/replays/__init__.py
TongTongX: done

hive/agents/qnets/rnn.py
TongTongX: If we do so we would have two `forward()` functions with a big chunk of shared code. Also we need the hidden state to compute the Q values for both act and update. There is no extra computation involved to compute the new hidden states. Thus for now it's fine if we use the same function for both act and updatae.

hive/agents/drqn.py
TongTongX: We have another PR https://github.com/chandar-lab/RLHive/pull/270 that handles hidden states saving & burn-in frames. In this PR the hidden states are initialized from 0's.

Could you also provide some reference if you have seen more efficient ways of reusing hidden state and computing Q?

dapatil211 (approved):
Assuming everything runs and matches existing benchmarks, looks good. Please add wandb runs to this conversation and also to the benchmarking report. Also when merging, do a squash and merge.

hive/agents/drqn.py
sriyash421: we can have a look at the cleanrl/sb3 implementation of recurrent networks. there implementation with jax might have some principles or tricks which we can use for our code base?

hive/agents/qnets/rnn.py
sriyash421: What you can do is have the act function modify the input and then pass it to forward? Just looks like a cleaner way to separate two use cases instead of having an if condition.

hive/agents/qnets/sequence_models.py
sriyash421: I dont understand. Isnt the hidden state part of the network module itself: the rnn layers etc.. and when you give the hidden state during forward pass you can use the device arg same as how we handle observation inputs

hive/replays/recurrent_replay.py
sriyash421: Can something like super calls be done? For e.g. the sample function in PPOReplayBuffer? That is the input is updated such that it can be passed to the function of the master class.

hnekoeiq:
> Assuming everything runs and matches existing benchmarks, looks good. Please add wandb runs to this conversation and also to the benchmarking report. Also when merging, do a squash and merge.

We should rerun the experiments for benchmarking, but here are some of the results on atari: 
https://github.com/chandar-lab/RLHive/pull/258#issuecomment-1075741637

And results on Hanabi DRQN: https://wandb.ai/chandar-rl/Hive/reports/DRQN-for-Hanabi--VmlldzoyMzU1NTUy?accessToken=efz63017kwgubdp8oyloepsfpjn3pmf0qka070h6rh90jjrcxcp8jd3eo20ko9sj

hive/agents/drqn.py
hnekoeiq: I think the current implementation is good enough and is working. We can create new prs to improve its efficiency maybe.

hive/agents/qnets/sequence_models.py
TongTongX: This is fixed by the latest commit, which update the device attribute of the sequence model once we move the whole network to cuda.

hive/agents/qnets/rnn.py
TongTongX: done

hive/replays/recurrent_replay.py
TongTongX: There will be a new PR that implements a recurrent replay that saves trajectories rather than sequences of transitions. It will be more suitable for recurrent DQN. For now I think it's fine to keep those functions where `max_seq_len` and `stack_size` are differerent.

sriyash421 (approved):

PR #272, opened 2022-05-03
PPO

### LINKED ISSUE #253: Add PPO Agent

### LINKED ISSUE #271: Buffer for on-policy algorithms like PPO, TRPO
Implementation of on policy algorithms would need an online buffer that holds the transitions, calculates the generalized advantage returns and then samples batches during training.

Or, should we leave this part on the agent side and have the agent calculate the returns.

This is needed because the current simple buffer samples random batches so we need additional features for on-policy learning.

kshitijkg:
PR that implements:

- PPO
- Normalization Wrappers
- Gym Wrappers


hive/agents/ppo.py
```diff
@@ -0,0 +1,307 @@
...
+        
+        if update_info["done"]:
+            self._state["episode_start"] = True
+
+        if not self._training:
+            return
+        
+        # Inspired  by https://github.com/vwxyzjn/ppo-implementation-details/blob/main/ppo_shared.py
+        # Check if the buffer is already full, if it is use current observation to bootstrap value estimate
+        if (self._replay_buffer.size() >= self._transitions_per_update): #We need next observation
```
sriyash421: For PPO I feel we can keep replay_buffer and transitions_per_update the same. We would need something like batch_iterations to reuse the entire replay buffer and then discard it. We don't need a circular history right?

Taking inspiration from https://github.com/fabiopardo/tonic/blob/master/tonic/torch/agents/ppo.py

@kshitijkg @dapatil211

kshitijkg: We want to perform minibatch gradient descent. So we want to use a minibatch of size batch_size for each update. 

We don't really need a circular history. But I have used that because that's the main replay buffer being used for other algorithms in RLHive. 

Following this blog: https://iclr-blog-track.github.io/2022/03/25/ppo-implementation-details/

hive/agents/ppo.py
```diff
@@ -0,0 +1,421 @@
...
+    calculate_output_dim,
+    create_init_weights_fn,
+)
+from hive.replays.ppo_replay import PPOReplayBuffer
+from hive.utils.loggers import Logger, NullLogger
+from hive.utils.schedule import (
+    PeriodicSchedule,
+)
+from hive.utils.utils import LossFn, OptimizerFn, create_folder
+from hive.agents.qnets.ppo_nets import PPOActorCriticNetwork, PPOActorNetwork, PPOCriticNetwork
```
saikrishna-1996: could you add `PPOActorCriticNetwork` to `ppo_nets`

hive/agents/qnets/ppo_nets.py
```diff
@@ -0,0 +1,123 @@
+import gym
+import numpy as np
+import torch
+
+from hive.agents.qnets.base import FunctionApproximator
+from hive.agents.qnets.utils import calculate_output_dim
+
+class PPOCriticNetwork(torch.nn.Module):
```
kshitijkg: Lets include an option to use a shared Actor Critic network and a Non shared actor critic Network?

Lets also include a config with the networks in 
https://iclr-blog-track.github.io/2022/03/25/ppo-implementation-details/

hive/agents/ppo.py
```diff
@@ -0,0 +1,411 @@
...
+
+        Args:
+            observation: The current observation.
+        """
+
+        observation = torch.tensor(
+            np.expand_dims(observation, axis=0), device=self._device
+        ).float()
+        #TODO: can add random exploration
+        action, logprob, _ = self._actor(observation)
```
kshitijkg: If representation is shared, there is double computation? Lets include a shared actor critic network to avoid that

hive/agents/ppo.py
```diff
@@ -0,0 +1,411 @@
...
+        """
+        super().__init__(observation_space, action_space, id)
+        self._device = torch.device("cpu" if not torch.cuda.is_available() else device)
+        self._state_size = (
+            stack_size * self._observation_space.shape[0],
+            *self._observation_space.shape[1:],
+        )
+        self._continuous_action = isinstance(self._action_space, gym.spaces.Box)
+        self._scale_actions = self._continuous_action
+        if self._scale_actions:
```
kshitijkg: Recheck scaling compared to other implementations

hive/agents/ppo.py
```diff
@@ -0,0 +1,411 @@
...
+            update_info: Contains the information from the current timestep that the
+                agent should use to update itself.
+        """
+        if self._reward_clip is not None:
+            update_info["reward"] = np.clip(
+                update_info["reward"], -self._reward_clip, self._reward_clip
+            )
+        preprocessed_update_info = {
+            "observation": update_info["observation"],
+            "action": self.scale_action(update_info["action"]),
```
kshitijkg: Why is this called scaling and not unscaling?

hive/agents/ppo.py
```diff
@@ -0,0 +1,411 @@
...
+        Args:
+            update_info: dictionary containing all the necessary information to
+                update the agent. Should contain a full transition, with keys for
+                "observation", "action", "reward", and "done".
+        """
+        if not self._training:
+            return
+        
+        # Inspired  by https://github.com/vwxyzjn/ppo-implementation-details/blob/main/ppo_shared.py
+        if (self._replay_buffer.ready()):
```
kshitijkg: Let's call this something else?

hive/agents/ppo.py
```diff
@@ -0,0 +1,411 @@
...
+                        self._logger.log_scalar("critic_loss", critic_loss, self._timescale)
+                        self._logger.log_scalar("entropy_loss", entr_loss, self._timescale)
+                        self._logger.log_scalar("approxkl", approx_kl, self._timescale)
+                
+                if self._target_kl is not None:
+                    if approx_kl > self._target_kl:
+                        break
+            self._replay_buffer.reset()
+
+        processed_update_info = self.preprocess_update_info(update_info)
```
kshitijkg: This is being done to get the next_obs to bootstrap the value estimte to calculate advantages. Issue with this approach is a transition from an older policy is used for the next update. 

@dapatil211, would it be possible to get the next_obs in the update info from the single agent runner loop?

hive/agents/ppo.py
```diff
@@ -0,0 +1,411 @@
...
+        # Inspired  by https://github.com/vwxyzjn/ppo-implementation-details/blob/main/ppo_shared.py
+        if (self._replay_buffer.ready()):
+            self._replay_buffer.compute_advantages(self._value)
+            #TODO: recheck sampling
+            for _ in range(self._num_epochs):
+                valid_ind_size = self._replay_buffer._find_valid_indices()
+                for _ in range(valid_ind_size//self._batch_size):
+                    batch = self._replay_buffer.sample(batch_size=self._batch_size)
+                    batch = self.preprocess_update_batch(batch)
+                    self._actor_optimizer.zero_grad()
```
kshitijkg: If we dont have one shared architecture, we have to deal with repeating actor and critic commands everytime

hive/agents/ppo.py
```diff
@@ -0,0 +1,411 @@
...
+                        critic_loss = 0.5 * v_loss_max.mean()
+                    else:
+                        critic_loss = 0.5 * self._critic_loss_fn(values, batch["returns"]).mean()
+                    critic_loss = self._vf_coef * critic_loss
+                    critic_loss.backward()
+                    if self._grad_clip is not None:
+                        torch.nn.utils.clip_grad_norm_(
+                            self._critic.parameters(), self._grad_clip
+                        )
+                    self._critic_optimizer.step()
```
kshitijkg: When you optimize the actor, the gradient is being backpropogated through the representation network making the current value loss inconsistent. This will result in incorrect loss.

hive/replays/ppo_replay.py
```diff
@@ -0,0 +1,104 @@
...
+            reward_dtype: Type of rewards that will be stored in the buffer. Format is
+                described in the description of observation_dtype.
+            extra_storage_types (dict): A dictionary describing extra items to store
+                in the buffer. The mapping should be from the name of the item to a
+                (type, shape) tuple.
+            num_players_sharing_buffer (int): Number of agents that share their
+                buffers. It is used for self-play.
+        """
+        #TODO: remove this
+        super().__init__(transitions_per_update+1, stack_size, n_step, gamma, observation_shape, observation_dtype, action_shape, action_dtype, reward_shape, reward_dtype, extra_storage_types, num_players_sharing_buffer)
```
kshitijkg: Lets not create a buffer with transitions_per_update+1

kshitijkg: we should instead use the normal transition_per_update and design accordingly

hive/replays/ppo_replay.py
```diff
@@ -0,0 +1,104 @@
...
+        self._sample_cursor = 0
+        self._transitions_per_update = transitions_per_update
+    
+    # Taken from https://github.com/vwxyzjn/ppo-implementation-details/blob/main/ppo_shared.py
+    def compute_advantages(self, values):
+        """ Compute advantages using rewards and value estimates.
+        """
+        if self._use_gae:
+            lastgaelam = 0
+            for t in reversed(range(self.size())):
```
kshitijkg: @dapatil211, curious why size returns capacity - 1 when buffer is full?

kshitijkg: @sriyash421 informed this is the reason we might need to use capacity variable?

hive/replays/ppo_replay.py
```diff
@@ -0,0 +1,104 @@
...
+                nextvalues = values if self.size()-1 else self._storage["values"][t + 1]
```
kshitijkg: t == self.size()-1 

Also should it be t == self.capacity -1?

hive/agents/ppo.py
kshitijkg: @sriyash421, this is still an issue that needs to be addressed.

@dapatil211, lets discuss this further

hive/agents/qnets/ppo_nets.py
kshitijkg: @sriyash421, could you include the shared actor-critic architecture from the original commit here as well?

sriyash421:
Here are the benchmarks
https://wandb.ai/sriyash421/Hive/reports/PPO-Benchmark-Results--VmlldzoyMTY1MzM3
https://wandb.ai/sriyash421/Hive-PPO-Benchmark/reports/PPO-Results-ICLR-blog-configs--VmlldzoyMDQzODQ2

hive/agents/qnets/ppo_nets.py
```diff
@@ -0,0 +1,139 @@
...
+            critic_network,
+            torch.nn.Flatten(),
+            torch.nn.Linear(feature_dim, 1),
+        )
+
+    def forward(self, obs):
+        return self.critic(obs)
+
+
+# TODO: add Multi-Discrete
```
kshitijkg: TODO?

hive/agents/ppo.py
kshitijkg: @sriyash421

dapatil211:
> Here are the benchmarks https://wandb.ai/sriyash421/Hive/reports/PPO-Benchmark-Results--VmlldzoyMTY1MzM3 https://wandb.ai/sriyash421/Hive-PPO-Benchmark/reports/PPO-Results-ICLR-blog-configs--VmlldzoyMDQzODQ2

Btw, I don't have access to these reports.

sriyash421:
@kshitijkg  I have updated the code with the shared architecture

sriyash421:
@dapatil211 @kshitijkg @chandar-lab/hive 
Latest benchmarks for Mujoco with performance comparisons: https://wandb.ai/sriyash-mila/Hive/reports/Mujoco--VmlldzoyMzEyMzk0?accessToken=nzylrw9kdo1spjxn0z6z3p3dgwfd75kwk6dpyuiyha2ipa468sone1n9pdbuj03q

Currently using out of the box wrappers from gym. TODO: extend these wrapper cleanly for all hive agents instead of using other codebase

sriyash421:
PS: Need reviews to clean up code and suggestions for improving design decisions (as some features were added to get a working codebase)

hive/envs/gym_env.py
```diff
@@ -29,7 +29,25 @@ def create_env(self, env_name, **kwargs):
...
+            env = gym.wrappers.TransformReward(
+                env, lambda reward: np.clip(reward, -10, 10)
+            )
+
+        elif kwargs.get("atari_wrapper", False):
+            env = gym.wrappers.NoopResetEnv(env, noop_max=30)
+            env = gym.wrappers.EpisodicLifeEnv(env)
+            if "FIRE" in env.unwrapped.get_action_meanings():
+                env = gym.wrappers.FireResetEnv(env)
+        self._env = env
```
dapatil211: This should be done differently. We should just register wrappers and rely on the users to provide a list of wrappers in the config.

hive/replays/circular_replay.py
```diff
@@ -242,7 +242,7 @@ def sample(self, batch_size):
 
         if self._n_step == 1:
             is_terminal = terminals
-            trajectory_lengths = np.ones(batch_size)
+            trajectory_lengths = np.ones(terminals.shape[0])
```
dapatil211: Why was this change made?

hive/agents/__init__.py
```diff
@@ -13,6 +14,7 @@
     {
         "DDPG": DDPG,
         "DQNAgent": DQNAgent,
+        "PPOAgent": PPOAgent,
```
dapatil211: Alphabetical order please.

hive/runners/base.py
```diff
@@ -86,7 +86,11 @@ def create_episode_metrics(self):
         return Metrics(
             self._agents,
             [("reward", 0), ("episode_length", 0)],
-            [("full_episode_length", 0)],
+            [
+                ("full_episode_length", 0),
+                ("episodic_length", 0),
+                ("episodic_return", 0),
```
dapatil211: what are these metrics? How are they different than what we have?

hive/agents/qnets/ppo_nets.py
```diff
@@ -0,0 +1,113 @@
...
+        _std = self.policy_logstd.repeat(x.shape[0], 1).exp()
+        distribution = self.distribution(
+            torch.reshape(_mean, (x.size(0), *self._action_shape)),
+            torch.reshape(_std, (x.size(0), *self._action_shape)),
+        )
+        return distribution
+
+
+class PPOActorCriticNetwork(torch.nn.Module):
+    """A module that implements the TD3 actor computation. It puts together the
```
dapatil211: Update docstring

hive/agents/ppo.py
```diff
@@ -0,0 +1,384 @@
...
+            optimizer_fn = torch.optim.Adam
+        self._optimizer = optimizer_fn(self._actor_critic.parameters())
+        if replay_buffer is None:
+            replay_buffer = PPOReplayBuffer
+        extra_storage_types = {
+            "values": (np.float32, ()),
+            "returns": (np.float32, ()),
+            "advantages": (np.float32, ()),
+            "logprob": (np.float32, ()),
+        }
```
dapatil211: 1. If you're creating a new buffer class for PPO anyway, this should be put inside the init function there.
2. You need to check if this dictionary has been passed, in case people want to put other extra stuff in the buffer.

hive/agents/qnets/ppo_nets.py
```diff
@@ -0,0 +1,113 @@
...
+    :py:class:`~torch.nn.Linear` layer to compute the action."""
+
+    def __init__(
+        self,
+        representation_network,
+        actor_net,
+        critic_net,
+        network_output_dim,
+        action_space,
+        continuous_action,
```
dapatil211: Type annotations please.

hive/agents/ppo.py
```diff
@@ -0,0 +1,384 @@
...
+            self._replay_buffer.compute_advantages(self._value)
+            for _ in range(self._num_epochs):
+                valid_ind_size = self._replay_buffer._find_valid_indices()
+                num_batches = int(np.ceil(valid_ind_size / self._batch_size))
+                for _ in range(num_batches):
+                    batch = self._replay_buffer.sample(batch_size=self._batch_size)
+                    batch = self.preprocess_update_batch(batch)
+                    self._optimizer.zero_grad()
+
+                    _, _logprob, entropy, values = self._actor_critic(
```
dapatil211: Why is there an underscore?

hive/agents/ppo.py
```diff
@@ -0,0 +1,384 @@
...
+                compared to the critic. The actor will be updated every
+                policy_update_frequency times the critic is updated.
+            action_noise (float): The standard deviation for the noise added to the
+                action taken by the agent during training.
+            target_noise (float): The standard deviation of the noise added to the
+                target policy for smoothing.
+            target_noise_clip (float): The sampled target_noise is clipped to
+                [-target_noise_clip, target_noise_clip].
+            min_replay_history (int): How many observations to fill the replay buffer
+                with before starting to learn.
```
dapatil211: Please update documentation

hive/agents/ppo.py
```diff
@@ -0,0 +1,384 @@
...
+        """
+        if not self._training:
+            return
+        if self._replay_buffer.ready():
+            self._replay_buffer.compute_advantages(self._value)
+            for _ in range(self._num_epochs):
+                valid_ind_size = self._replay_buffer._find_valid_indices()
+                num_batches = int(np.ceil(valid_ind_size / self._batch_size))
+                for _ in range(num_batches):
+                    batch = self._replay_buffer.sample(batch_size=self._batch_size)
```
dapatil211: I think this should maybe be rewritten as follows:
```
if self._replay_buffer.size() + 1 == self._transitions_per_update: # You can make this an agent parameter instead of a buffer parameter.
    batches = self._replay_buffer.sample(self._batch_size) # This would compute advantages and return a list of batches
    for batch in batches:
        ...
```

This avoids adding too many methods to the buffer, and keeps the api a bit simpler. Also you aren't accessing internal methods of the buffer. Let me know if this doesn't work for some reason.

hive/replays/ppo_replay.py
```diff
@@ -0,0 +1,149 @@
+import numpy as np
+from hive.replays.circular_replay import CircularReplayBuffer
+
+
+class PPOReplayBuffer(CircularReplayBuffer):
```
dapatil211: See my comments in the ppo.py file.

hive/agents/ppo.py
```diff
@@ -0,0 +1,384 @@
...
+                    logratios = _logprob - batch["logprob"]
+                    ratios = torch.exp(logratios)
+                    advantages = batch["advantages"]
+                    if self._normalize_advantages:
+                        advantages = (advantages - advantages.mean()) / (
+                            advantages.std() + 1e-8
+                        )
+                    # Actor loss
+                    loss1 = -advantages * ratios
+                    loss2 = -advantages * torch.clamp(
```
dapatil211: Is it possible to give these more descriptive names?

hive/agents/ppo.py
```diff
@@ -0,0 +1,384 @@
...
+
+        processed_update_info = self.preprocess_update_info(update_info)
+        processed_update_info.update(
+            {
+                "logprob": self._logprob,
+                "values": self._value,
+                "returns": np.empty(self._value.shape),
+                "advantages": np.empty(self._value.shape),
+            }
+        )
```
dapatil211: Can't you do this inside of preprocess_update_info?

dapatil211 (changes requested):

hive/replays/circular_replay.py
sriyash421: for the ppo agent, in case of stack_size > 1, if we have a episode reset in between the replay, then the new padded transitions are not valid indices yet they contribute to the size. So it leads to the last batch with transitions less than batch_size, and this part throws error.

hive/envs/gym_env.py
sriyash421: I will do that but the most commonly used wrappers are the normalizers and clipping wrappers which I was thinking to code as a part of hive and include it in the function approximator pipeline

sriyash421: or do i register all the wrappers and use them as it is

hive/replays/circular_replay.py
dapatil211: Ok, I think if you change the api as i mention in ppo.py, you'll need to override sample anyway, in which case this change wouldn't be necessary. Let me know if you have any comments on the proposed changes.

hive/agents/ppo.py
sriyash421: _logprob is the new logprob of the action. I'll change the name as _ represents internal methods and variables right?

hive/agents/ppo.py
sriyash421: Yes moved the function

hive/agents/ppo.py
sriyash421: yes named them as clipped and unclipped policy losses

hive/agents/ppo.py
sriyash421: I will try and use this idea and see if it works!

hive/agents/ppo.py
sriyash421: I moved this function to the new buffer and update the extra_args passed by the user

hive/runners/base.py
sriyash421: I was using the normalising wrappers for reward which returned the normalised rewards each step and that is what was logged. Common practice was to use RecordEpisodeStats wrapper along with it which returns the original reward and ep length at the end of the episode, so these metrics capture that. But I plan to make changes to the api having our own funnctions inside the agents we can remove them then

hive/envs/gym_env.py
dapatil211: I think for now you can just register the wrappers and use them as is.

hive/runners/base.py
sriyash421: @dapatil211  any comments on this?

dapatil211: Yeah I'd say if you need to normalize the reward, you can do it in preprocess_update_info like its done in dqn. I'd also get rid of all the hardcoded wrappers that you have.

hive/agents/ppo.py
kshitijkg: @dapatil211, let me know what you think about the new code. 

We could potentially move compute advantages inside the buffer and change the sample function to include the next_value. The issue is that it changes the api of the sample function and also recomputes advantages at every epoch, which is not necessary for now and decreases the throughput of the agent. **Though in my previous experience that helps prevent stale hidden states when using a recurrent poliy.

hive/envs/base.py
```diff
@@ -158,3 +160,44 @@ def step(self, action):
...
+    files or command line arguments.
+    """
+
+    @classmethod
+    def type_name(cls):
+        """
+        Returns:
+            "init_fn"
+        """
+        return "init_fn"
```
dapatil211: type name should be something else

hive/envs/base.py
```diff
@@ -158,3 +160,44 @@ def step(self, action):
...
+    EnvWrapper,
+    {
+        "RecordEpisodeStatistics": gym.wrappers.RecordEpisodeStatistics,
+        "ClipAction": gym.wrappers.ClipAction,
+        "NormalizeObservation": gym.wrappers.NormalizeObservation,
+        "TransformObserva": gym.wrappers.TransformObservation,
+        "NormalizeReward": gym.wrappers.NormalizeReward,
+        "TransformReward": gym.wrappers.TransformReward,
+        "CompositeEnvWrapper": CompositeEnvWrapper,
+    },
```
dapatil211: Since these are only gym wrappers, you should maybe register them in the gym file.

hive/replays/ppo_replay.py
```diff
@@ -0,0 +1,158 @@
...
+    ):
+        """Constructor for PPOReplayBuffer.
+
+        Args:
+            capacity (int): Total number of observations that can be stored in the buffer
+            stack_size (int): The number of frames to stack to create an observation.
+            n_step (int): Horizon used to compute n-step return reward
+            gamma (float): Discounting factor used to compute n-step return reward
+            use_gae (bool): Whether to use generalised advantage estimates for calculating returns
+            gae_lambda (float): Discouting factor used to compute generalised advantage estimation
```
dapatil211: Please make sure lines are less than 88 characters

hive/replays/ppo_replay.py
```diff
@@ -0,0 +1,158 @@
...
+            num_players_sharing_buffer,
+        )
+        self._use_gae = use_gae
+        self._gae_lambda = gae_lambda
+
+    # Taken from https://github.com/vwxyzjn/ppo-implementation-details/blob/main/ppo_shared.py
+    def compute_advantages(self, values):
+        """Compute advantages using rewards and value estimates."""
+        if self._use_gae:
+            lastgaelam = 0
```
dapatil211: Variable names with more than 1 word should have underscore eg last_gae_lam. Also when possible, please use full names to avoid confusion, eg last_gae_lambda. Please also go through and fix the other instances.

hive/runners/single_agent_loop.py
```diff
@@ -84,6 +85,9 @@ def run_one_step(self, observation, episode_metrics):
         episode_metrics[agent.id]["reward"] += info["reward"]
         episode_metrics[agent.id]["episode_length"] += 1
         episode_metrics["full_episode_length"] += 1
+        if "episode" in other_info.keys():
+            episode_metrics["episodic_return"] += other_info["episode"]["r"]
+            episode_metrics["episodic_length"] += other_info["episode"]["l"]
```
dapatil211: This isn't great. I am guessing the only reason this is needed is because of the normalize reward? If that's the case, can that normalization be done inside the agent instead of an environment wrapper? It feels more like the responsibility of the agent anyways.

hive/agents/ppo.py
```diff
@@ -0,0 +1,379 @@
...
+    @torch.no_grad()
+    def act(self, observation):
+        """Returns the action for the agent.
+
+        Args:
+            observation: The current observation.
+        """
+        action, logprob, value = self.get_action_logprob_value(observation)
+        self._logprob = logprob
+        self._value = value
```
dapatil211: Could you put this into a single state dictionary? It will be easier to port when I add the changes for parallel environments, as the agents will become a bit more stateless (they won't store these state variables, instead they will pass those dictionaries around)

hive/agents/ppo.py
```diff
@@ -0,0 +1,379 @@
...
+)
+from hive.replays.ppo_replay import PPOReplayBuffer
+from hive.utils.loggers import Logger, NullLogger
+from hive.utils.schedule import (
+    PeriodicSchedule,
+)
+from hive.utils.utils import LossFn, OptimizerFn, create_folder
+
+
+class PPOAgent(Agent):
```
dapatil211, line 26: Any idea how fast this runs compared to other libraries with single environments?

dapatil211 (changes requested):

hive/configs/gym/ppo.yml
```diff
@@ -0,0 +1,64 @@
+run_name: &run_name 'ppo'
```
hnekoeiq: change the name to 'gym-ppo'?

hive/configs/mujoco/ppo.yml
```diff
@@ -0,0 +1,74 @@
+run_name: &run_name 'ppo'
```
hnekoeiq: 'mujoco-ppo'?

hnekoeiq (changes requested):

hive/replays/ppo_replay.py
```diff
@@ -0,0 +1,158 @@
...
+            action_dtype,
+            reward_shape,
+            reward_dtype,
+            extra_storage_types,
+            num_players_sharing_buffer,
+        )
+        self._use_gae = use_gae
+        self._gae_lambda = gae_lambda
+
+    # Taken from https://github.com/vwxyzjn/ppo-implementation-details/blob/main/ppo_shared.py
```
hnekoeiq: @dapatil211 Correct me if I'm wrong, but it would be better to get inspiration from others' code but write our own version. I know there aren't many ways to implement this function but we can probably at least reformat it.

hive/agents/ppo.py
```diff
@@ -0,0 +1,379 @@
...
+                observations from representation_net and outputs the representations
+                used to compute the actions (ie everything except the last layer).
+            critic_net (FunctionApproximator): The network that takes two inputs: the
+                encoded observations from representation_net and actions. It outputs
+                the representations used to compute the values of the actions (ie
+                everything except the last layer).
+            init_fn (InitializationFn): Initializes the weights of agent networks using
+                create_init_weights_fn.
+            optimizer_fn (OptimizerFn): A function that takes in the list of
+                parameters of the actor and critic returns the optimizer for the actor. If None,
```
dapatil211: line is too long

hive/envs/gym_env.py
```diff
@@ -70,3 +87,44 @@ def seed(self, seed=None):
...
+    def __init__(self, env, wrapper_list: List[EnvWrapper]):
+        for wrapper in wrapper_list:
+            env = wrapper(env)
+        super().__init__(env)
+
+
+registry.register_all(
+    EnvWrapper,
+    {
+        "RecordEpisodeStatistics": gym.wrappers.RecordEpisodeStatistics,
```
kshitijkg: Remove

hive/envs/gym_env.py
```diff
@@ -70,3 +87,44 @@ def seed(self, seed=None):
...
+        "ClipAction": gym.wrappers.ClipAction,
+        "NormalizeObservation": gym.wrappers.NormalizeObservation,
+        "TransformObserva": gym.wrappers.TransformObservation,
+        "NormalizeReward": gym.wrappers.NormalizeReward,
+        "TransformReward": gym.wrappers.TransformReward,
```
kshitijkg: Where are these being used? Remove if not needed

hive/agents/qnets/normalizer.py
```diff
@@ -0,0 +1,150 @@
...
+    def __call__(self, rew):
+        rew = rew / np.sqrt(self.return_rms.var + self._epsilon)
+        if self._clip is not None:
+            rew = np.clip(rew, -self._clip, self._clip)
+        return rew
+
+    def update(self, rew, done):
+        self._returns = self._returns * self._gamma + rew
+        self.return_rms.update(self._returns)
+        self._returns *= 1 - done
```
kshitijkg: Leading dimension should repesent number of environemnts

hive/agents/ppo.py
```diff
@@ -0,0 +1,401 @@
...
+        observation = torch.tensor(
+            np.expand_dims(observation, axis=0), device=self._device
+        ).float()
+        action, logprob, _, value = self._actor_critic(observation)
+        action = action.cpu().detach().numpy()
+        logprob = logprob.cpu().numpy()
+        value = value.cpu().numpy()
+        action = action[0]
+
+        return action, logprob, value
```
kshitijkg, line 281: Not using normalized observation

hive/agents/ppo.py
```diff
@@ -0,0 +1,401 @@
...
+
+    def preprocess_update_info(self, update_info):
+        """Preprocesses the :obj:`update_info` before it goes into the replay buffer.
+
+        Args:
+            update_info: Contains the information from the current timestep that the
+                agent should use to update itself.
+        """
+        if self.obs_norm:
+            self.obs_norm.update(update_info["observation"])
```
kshitijkg: Observation normalization statistics will already be updated in the act function

hive/agents/ppo.py
```diff
@@ -0,0 +1,405 @@
...
+            init_fn (InitializationFn): Initializes the weights of agent networks using
+                create_init_weights_fn.
+            optimizer_fn (OptimizerFn): A function that takes in the list of
+                parameters of the actor and critic returns the optimizer for the actor. If None,
+                defaults to :py:class:`~torch.optim.Adam`.
+            critic_loss_fn (LossFn): The loss function used to optimize the critic. If
+                None, defaults to :py:class:`~torch.nn.MSELoss`.
+            observation_normalizer (FunctionApproximator): The function for normalizing
+                observations
+            reward_normalizer (FunctionApproximator): The function for normalizing rewards
```
dapatil211: Please make sure this documentation is accurate. Also line 82 is too long.

hive/agents/ppo.py
```diff
@@ -0,0 +1,405 @@
...
+            stack_size (int): Number of observations stacked to create the state fed
+                to the agent.
+            replay_buffer (BaseReplayBuffer): The replay buffer that the agent will
```
dapatil211: Please make sure this is accurate

hive/agents/ppo.py
```diff
@@ -0,0 +1,405 @@
...
+            clip_vloss (bool): Flag to use the clipped objective for the value
+                function.
+            vf_coef (float): Coefficient for the value function loss.
+            transitions_per_update (int): Total number of observations that are
+                stored before the update.
+            num_epochs_per_update (int): Number of iterations over the entire
+                buffer during an update step.
+            normalize_advantages (bool): Flag to normalise advantages before
+                calculating policy loss.
+            target_kl: Terminates the update if kl-divergence between old and
```
dapatil211: Please add type annotation

hive/agents/ppo.py
```diff
@@ -0,0 +1,405 @@
...
+                update the agent. Should contain a full transition, with keys for
+                "observation", "action", "reward", and "done".
+        """
+        if not self._training:
+            return
+
+        # Add the most recent transition to the replay buffer.
+        self._replay_buffer.add(**self.preprocess_update_info(update_info))
+
+        if self._replay_buffer.size() + 1 == self._transitions_per_update:
```
dapatil211: I'd maybe write this logic as 
```suggestion
        if self._replay_buffer.size() == self._transitions_per_update - 1:
```

hive/agents/ppo.py
```diff
@@ -0,0 +1,405 @@
...
+            if self.obs_norm:
+                update_info["next_observation"] = self.obs_norm(
+                    update_info["next_observation"]
+                )
+            self._replay_buffer.compute_advantages(
+                self.get_action_logprob_value(update_info["next_observation"])[2]
+            )
```
dapatil211: This is a little bit of a nitpick, but maybe write this as 
```suggestion
            _, _, values =  self.get_action_logprob_value(update_info["next_observation"])
            self._replay_buffer.compute_advantages(values)
```

hive/agents/ppo.py
```diff
@@ -0,0 +1,405 @@
...
+            device: Device on which all computations should be run.
+            id: Agent identifier.
+        """
+        super().__init__(observation_space, action_space, id)
+        self._device = torch.device("cpu" if not torch.cuda.is_available() else device)
+        self._state_size = (
+            stack_size * self._observation_space.shape[0],
+            *self._observation_space.shape[1:],
+        )
+        self._continuous_action = isinstance(self._action_space, gym.spaces.Box)
```
dapatil211: This probably doesn't need to be class variable, you can just do this check where it's needed.

hive/agents/ppo.py
```diff
@@ -0,0 +1,405 @@
...
+        )
+        if obs_norm is not None:
+            self.obs_norm = obs_norm(self._state_size)
+        else:
+            self.obs_norm = None
+
+        if rew_norm is not None:
+            self.rew_norm = rew_norm(discount_rate)
+        else:
+            self.rew_norm = None
```
dapatil211: Can you expand these variable names? Also if these are supposed to be private/not accessed from outside, put and underscore in front.

hive/agents/ppo.py
```diff
@@ -0,0 +1,405 @@
...
+        if self._logger is None:
+            self._logger = NullLogger([])
+        self._timescale = self.id
+        self._logger.register_timescale(
+            self._timescale, PeriodicSchedule(False, True, log_frequency)
+        )
+        self._clip_coef = clip_coef
+        self._ent_coef = ent_coef
+        self._clip_vloss = clip_vloss
+        self._vf_coef = vf_coef
```
dapatil211: I'd also expand these variable names

hive/agents/ppo.py
```diff
@@ -0,0 +1,405 @@
...
+        return preprocessed_update_info
+
+    def preprocess_update_batch(self, batch):
+        """Preprocess the batch sampled from the replay buffer.
+
+        Args:
+            batch: Batch sampled from the replay buffer for the current update.
+
+        Returns:
+                - Preprocessed batch.
```
dapatil211: fix indentation

hive/agents/ppo.py
```diff
@@ -0,0 +1,405 @@
...
+        """Returns the action for the agent.
+
+        Args:
+            observation: The current observation.
+        """
+        if self.obs_norm:
+            self.obs_norm.update(observation)
+            observation = self.obs_norm(observation)
+        action, logprob, value = self.get_action_logprob_value(observation)
+        self._act_vars = dict(logprob=logprob, value=value)
```
dapatil211: If this is a state dictionary, could you modify the code to be in the format of the stateless_agent code? That will likely get merged before this pr.

hive/agents/qnets/utils.py
```diff
@@ -1,5 +1,7 @@
 import math
+from typing import Tuple
 
+import numpy as np
```
dapatil211: Any reason for these changes?

hive/envs/gym_env.py
```diff
@@ -29,7 +33,19 @@ def create_env(self, env_name, **kwargs):
...
+
+        wrapper_config = kwargs.get("wrappers", None)
+        if isinstance(wrapper_config, list) and len(wrapper_config) > 0:
+            wrapper_config = {
+                "name": "CompositeEnvWrapper",
+                "kwargs": {"wrapper_list": wrapper_config},
+            }
+        if wrapper_config is not None:
+            wrapper_fn, _ = get_wrapper(wrapper_config, "wrappers")
+            env = wrapper_fn(env)
```
dapatil211: This isn't super clean, CompositeEnvWrapper isn't really necessary. Something better would be add a parameter to the init called `env_wrappers: List[Wrapper]`. Then below:
```suggestion
def create_env(self, env_name, env_wrappers, **kwargs):
        self._env = gym.make(env_name)
        
        if env_wrappers is not None:
            self._env = apply_wrappers(env, env_wrappers)

def apply_wrappers(env, env_wrappers):
    for wrapper in env_wrappers:
        env = wrapper(env)
    return env         
```

`apply_wrappers` and the EnvWrapper definition should be moved to their own file. This way, the wrapper logic is also all gym independent.

hive/replays/ppo_replay.py
```diff
@@ -0,0 +1,163 @@
+import numpy as np
+from hive.replays.circular_replay import CircularReplayBuffer
+
+
+class PPOReplayBuffer(CircularReplayBuffer):
```
dapatil211: Maybe rename this OnPolicyReplayBuffer?

hive/replays/ppo_replay.py
```diff
@@ -0,0 +1,163 @@
...
+            reward_shape,
+            reward_dtype,
+            extra_storage_types,
+            num_players_sharing_buffer,
+        )
+        self._use_gae = use_gae
+        self._gae_lambda = gae_lambda
+
+    # Taken from https://github.com/vwxyzjn/ppo-implementation-details/blob/main/ppo_shared.py
+    def compute_advantages(self, values):
```
dapatil211: Instead of this, can you maybe just have the buffer take in a (registered) function as an argument that will then compute the advantages? That way, if someone wants to compute advantages in a different way, they can use the same replay, just pass a different function for computing the advantages. If you do this, you should have this function just take in the rewards and values and return advantages/returns. 
I am also not sure why this computation needs to be done in the replay buffer? Can you not just do it in the agent, and then use the indices returned in the batch. Not sure either way, but that might be a better division of responsibilities. Let me know if you have an argument for the other way.

hive/replays/ppo_replay.py
```diff
@@ -0,0 +1,163 @@
...
+                    + self._gamma * next_non_terminal * next_return
+                )
+            self._storage["advantages"] = (
+                self._storage["returns"] - self._storage["values"]
+            )
+
+    def reset(self):
+        """Resets the storage."""
+        if self._stack_size > 1:
+            transitions = {
```
dapatil211: Maybe give a better name, something like saved_transitions? Or something else, so that the code is a bit more self-documenting.

hive/replays/ppo_replay.py
```diff
@@ -0,0 +1,163 @@
...
+                k: self._storage[k][-(self._stack_size - 1) :]
+                for k in self._storage.keys()
+            }
+            self._create_storage(self._capacity, self._specs)
+            for k in self._storage.keys():
+                self._storage[k][: (self._stack_size - 1)] = transitions[k]
+            self._episode_start = transitions["done"][-1]
+        else:
+            self._create_storage(self._capacity, self._specs)
+            self._episode_start = True
```
dapatil211: Where is this used?

dapatil211 (changes requested):

hive/agents/ppo.py
```diff
@@ -0,0 +1,413 @@
...
+        Args:
+            observation: The current observation.
+        """
+        if agent_traj_state is None:
+            agent_traj_state = {}
+        if self._observation_normalization_fn:
+            self._observation_normalization_fn.update(observation)
+            observation = self._observation_normalization_fn(observation)
+        action, logprob, value = self.get_action_logprob_value(observation)
+        agent_traj_state["act_vars"] = dict(logprob=logprob, value=value)
```
dapatil211: You probably don't need to put this inside of a subdictionary

hive/agents/ppo.py
```diff
@@ -0,0 +1,413 @@
...
+        """
+        if not self._training:
+            return
+
+        # Add the most recent transition to the replay buffer.
+        self._replay_buffer.add(
+            **self.preprocess_update_info(update_info, agent_traj_state)
+        )
+
+        if self._replay_buffer.size() == self._transitions_per_update - 1:
```
dapatil211: should this be == or >=? The latter feels more robust to changes in the future, but maybe there is no case where > would be hit

hive/agents/ppo.py
```diff
@@ -0,0 +1,413 @@
...
+            if self._observation_normalization_fn:
+                update_info["next_observation"] = self._observation_normalization_fn(
+                    update_info["next_observation"]
+                )
+            _, _, values = self.get_action_logprob_value(
+                update_info["next_observation"]
+            )
```
dapatil211: This might be more readable:
```suggestion
            if self._observation_normalization_fn:
                next_observation = self._observation_normalization_fn(
                    update_info["next_observation"]
                )
            _, _, values = self.get_action_logprob_value(
                next_observation
            )
```

hive/agents/ppo.py
```diff
@@ -0,0 +1,413 @@
...
+                        self._logger.log_scalar(
+                            "actor_loss", actor_loss, self._timescale
+                        )
+                        self._logger.log_scalar(
+                            "critic_loss", critic_loss, self._timescale
+                        )
+                        self._logger.log_scalar(
+                            "entropy_loss", entr_loss, self._timescale
+                        )
+                        self._logger.log_scalar("approxkl", approx_kl, self._timescale)
```
dapatil211: missing underscore

hive/agents/qnets/normalizer.py
```diff
@@ -0,0 +1,151 @@
...
+    def update_from_moments(self, batch_mean, batch_var, batch_count):
+        """Updates from batch mean, variance and count moments."""
+        self.mean, self.var, self.count = self.update_mean_var_count_from_moments(
+            self.mean, self.var, self.count, batch_mean, batch_var, batch_count
+        )
+
+    def update_mean_var_count_from_moments(
+        self, mean, var, count, batch_mean, batch_var, batch_count
+    ):
+        """Updates the mean, var and count using the previous mean, var, count and batch values."""
```
dapatil211: line length

hive/agents/qnets/ppo_nets.py
```diff
@@ -0,0 +1,116 @@
...
+import numpy as np
+import torch
+
+from hive.agents.qnets.base import FunctionApproximator
+from hive.agents.qnets.utils import calculate_output_dim
+
+
+class CategoricalHead(torch.nn.Module):
+    """A module that implements a discrete actor head. It uses the ouput from the
+    :obj:`actor_net`, and adds creates a :py:class:`~torch.distributions.categorical.Categorical`
```
dapatil211: line length

hive/agents/qnets/ppo_nets.py
```diff
@@ -0,0 +1,116 @@
...
+            torch.reshape(_std, (x.size(0), *self._action_shape)),
+        )
+        return distribution
+
+
+class PPOActorCriticNetwork(torch.nn.Module):
+    """A module that implements the PPO actor and critic computation. It puts together the
+    :obj:`representation_network`, :obj:`actor_net` and :obj:`critic_net`, then adds two final
+    :py:class:`~torch.nn.Linear` layers to compute the action and state value."""
+
```
dapatil211: line length

hive/envs/env_wrapper.py
```diff
@@ -0,0 +1,38 @@
...
+    EnvWrapper,
+    {
+        "RecordEpisodeStatistics": gym.wrappers.RecordEpisodeStatistics,
+        "ClipAction": gym.wrappers.ClipAction,
+        "NormalizeObservation": gym.wrappers.NormalizeObservation,
+        "TransformObservation": gym.wrappers.TransformObservation,
+        "NormalizeReward": gym.wrappers.NormalizeReward,
+        "TransformReward": gym.wrappers.TransformReward,
+    },
+)
```
dapatil211: gym specific wrappers should be registered in the file with the gym env code. you should also register all the wrappers that gym supports

hive/envs/gym_env.py
```diff
@@ -1,37 +1,44 @@
...
 
 
 class GymEnv(BaseEnv):
     """
     Class for loading gym environments.
     """
 
-    def __init__(self, env_name, num_players=1, render_mode=None, **kwargs):
+    def __init__(
+        self, env_name, env_wrappers=None, num_players=1, render_mode=None, **kwargs
```
dapatil211: env_wrappers needs a type annotation of List[EnvWrapper]

hive/replays/on_policy_replay.py
```diff
@@ -0,0 +1,161 @@
...
+        num_players_sharing_buffer: int = None,
+    ):
+        """Constructor for OnPolicyReplayBuffer.
+
+        Args:
+            capacity (int): Total number of observations that can be stored in the buffer
+            stack_size (int): The number of frames to stack to create an observation.
+            n_step (int): Horizon used to compute n-step return reward
+            gamma (float): Discounting factor used to compute n-step return reward
+            use_gae (bool): Whether to use generalised advantage estimates for calculating
```
dapatil211: line length

hive/replays/on_policy_replay.py
```diff
@@ -0,0 +1,161 @@
...
+                    else self._storage["return"][t + 1]
+                )
+                next_non_terminal = 1.0 - self._storage["done"][t]
+                self._storage["returns"][t] = (
+                    self._storage["reward"][t]
+                    + self._gamma * next_non_terminal * next_return
+                )
+            self._storage["advantages"] = (
+                self._storage["returns"] - self._storage["values"]
+            )
```
dapatil211: Please extract these advantage calculations into two separate methods, one for gae and one for non gae.

hive/runners/single_agent_loop.py
```diff
@@ -115,6 +115,7 @@ def run_one_step(
 
         info = {
             "observation": observation,
+            "next_observation": next_observation,
```
dapatil211, line 118: please also make these changes in the multi agent runner

dapatil211 (changes requested):

hive/envs/gym_env.py
```diff
@@ -1,37 +1,48 @@
...
```
dapatil211: You need to add the type annotation to the signature

hive/envs/gym_env.py
```diff
@@ -1,37 +1,48 @@
 import gymnasium as gym
+import inspect
+import gymnasium.wrappers as w
```
dapatil211: Please run an import sorter

dapatil211: Also please refactor this as
`from gymnasium import wrappers`
to avoid single letter variables

dapatil211 (changes requested):
A couple of other changes

hive/agents/ppo.py
```diff
@@ -0,0 +1,412 @@
...
+            init_fn (InitializationFn): Initializes the weights of agent networks
+                using create_init_weights_fn.
+            optimizer_fn (OptimizerFn): A function that takes in the list of
+                parameters of the actor and critic returns the optimizer for the actor.
+                If None, defaults to :py:class:`~torch.optim.Adam`.
+            critic_loss_fn (LossFn): The loss function used to optimize the critic. If
+                None, defaults to :py:class:`~torch.nn.MSELoss`.
+            observation_normalizer (NormalizationFn): The function for normalizing
+                observations
+            reward_normalizer (NormalizationFn): The function for normalizing rewards
```
dapatil211: Make these consistent with actual names

hive/agents/ppo.py
```diff
@@ -0,0 +1,412 @@
...
+                        self._logger.log_scalar(
+                            "actor_loss", actor_loss, self._timescale
+                        )
+                        self._logger.log_scalar(
+                            "critic_loss", critic_loss, self._timescale
+                        )
+                        self._logger.log_scalar(
+                            "entropy_loss", entr_loss, self._timescale
+                        )
+                        self._logger.log_scalar("approx_kl", approx_kl, self._timescale)
```
dapatil211: you can log these together as one dictionary with `logger.log_metrics`

hive/agents/ppo.py
```diff
@@ -0,0 +1,412 @@
...
+
+                if self._target_kl is not None:
+                    if approx_kl > self._target_kl:
+                        break
```
dapatil211:
```suggestion
                if self._target_kl is not None and self._target_kl < approx_kl:
                    break
```

hive/agents/ppo.py
```diff
@@ -0,0 +1,412 @@
...
+        )
+        replay_dir = os.path.join(dname, "replay")
+        create_folder(replay_dir)
+        self._replay_buffer.save(replay_dir)
+
+    def load(self, dname):
+        checkpoint = torch.load(os.path.join(dname, "agent.pt"))
+        self._actor_critic.load_state_dict(checkpoint["actor_critic"])
+        self._optimizer.load_state_dict(checkpoint["optimizer"])
+        self._replay_buffer.load(os.path.join(dname, "replay"))
```
dapatil211: There's also state in the normalizers

hive/agents/qnets/normalizer.py
```diff
@@ -0,0 +1,153 @@
...
+        pass
+
+    def __call__(self, *args, **kwds):
+        return NotImplementedError
+
+    def update(self, *args, **kwds):
+        return NotImplementedError
+
+
+class ObservationNormalizationFn(BaseNormalizationFn):
```
dapatil211: Maybe rename to something like `MovingAvgNormalizer`, since it's not specific to observations. Also, maybe update variable names (obs) to reflect generality.

hive/agents/qnets/normalizer.py
```diff
@@ -0,0 +1,153 @@
...
+        m_a = var * count
+        m_b = batch_var * batch_count
+        M2 = m_a + m_b + np.square(delta) * count * batch_count / tot_count
+        new_var = M2 / tot_count
+        new_count = tot_count
+
+        return new_mean, new_var, new_count
+
+
+class BaseNormalizationFn(object):
```
dapatil211: Maybe rename to `Normalizer` since these aren't functions. Also, just directly subclass Registrable.

hive/agents/qnets/normalizer.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def update(self, obs):
+        self.obs_rms.update(obs)
+
+
+class RewardNormalizationFn(BaseNormalizationFn):
+    """Implements a normalization function. Transforms output by
+    normalising the input data by the running :obj:`mean` and
+    :obj:`std`, and clipping the normalised data on :obj:`clip`
+    """
```
dapatil211: This docstring seems a bit incomplete/inaccurate.

hive/agents/qnets/normalizer.py
```diff
@@ -0,0 +1,153 @@
...
+        obs = ((obs - self.obs_rms.mean) / np.sqrt(self.obs_rms.var + self._epsilon))[0]
+        if self._clip is not None:
+            obs = np.clip(obs, -self._clip, self._clip)
+        return obs
+
+    def update(self, obs):
+        self.obs_rms.update(obs)
+
+
+class RewardNormalizationFn(BaseNormalizationFn):
```
dapatil211: Maybe change to `RewardNormalizer`

hive/agents/qnets/normalizer.py
```diff
@@ -0,0 +1,153 @@
...
+            rew = np.clip(rew, -self._clip, self._clip)
+        return rew
+
+    def update(self, rew, done):
+        self._returns = self._returns * self._gamma + rew
+        self.return_rms.update(self._returns)
+        self._returns *= 1 - done
+
+
+class NormalizationFn(Registrable):
```
dapatil211: You don't need this and the other base class. Just use one, and subclass Registrable. Just provide the type_name

hive/agents/qnets/normalizer.py
```diff
@@ -0,0 +1,153 @@
...
+        Returns:
+            "norm_fn"
+        """
+        return "norm_fn"
+
+
+registry.register_all(
+    NormalizationFn,
+    {
+        "BaseNormalization": BaseNormalizationFn,
```
dapatil211: BaseNormalizationFn shouldn't be registered since you don't want someone to actually construct it. Just delete.

hive/agents/qnets/ppo_nets.py
```diff
@@ -0,0 +1,119 @@
...
+        _mean = self.policy_mean(x)
+        _std = self.policy_logstd.repeat(x.shape[0], 1).exp()
+        distribution = self.distribution(
+            torch.reshape(_mean, (x.size(0), *self._action_shape)),
+            torch.reshape(_std, (x.size(0), *self._action_shape)),
+        )
+        return distribution
+
+
+class PPOActorCriticNetwork(torch.nn.Module):
```
dapatil211: Is this specific to PPO? Or is it general to all/most AC methods

hive/envs/gym_env.py
```diff
@@ -1,37 +1,55 @@
...
-    def create_env(self, env_name, **kwargs):
+    def create_env(self, env_name, env_wrappers, **kwargs):
         """Function used to create the environment. Subclasses can override this method
         if they are using a gym style environment that needs special logic.
 
         Args:
             env_name (str): Name of the environment
         """
-        self._env = gym.make(env_name, **kwargs)
+        self._env = gym.make(env_name)
```
dapatil211: Please undo this change.

hive/replays/on_policy_replay.py
```diff
@@ -0,0 +1,131 @@
...
+        gamma: float = 0.99,
+        compute_advantage_fn: AdvantageComputationFn = None,
+        observation_shape=(),
+        observation_dtype=np.uint8,
+        action_shape=(),
+        action_dtype=np.int8,
+        reward_shape=(),
+        reward_dtype=np.float32,
+        extra_storage_types=None,
+        num_players_sharing_buffer: int = None,
```
dapatil211: Not sure this works with multiple agents. Maybe remove this from the parameter list.

dapatil211 (changes requested):

hive/agents/qnets/normalizer.py
```diff
@@ -46,24 +47,44 @@ def update_mean_var_count_from_moments(
...
+
+    @abc.abstractmethod
+    def state_dict(self):
+        """Returns the state of the normalizer as a dictionary."""
 
-class ObservationNormalizationFn(BaseNormalizationFn):
-    """Implements a normalization function. Transforms output by
-    normalising the input data by the running :obj:`mean` and
-    :obj:`std`, and clipping the normalised data on :obj:`clip`
+    def load_state_dict(self, state_dict):
```
dapatil211, line 81: This should also be abstract right?

dapatil211 (changes requested):

PR #275, opened 2022-05-03
Ddpg + TD3

### LINKED ISSUE #254: Add DDPG, TD3 Agents

dapatil211:
Experiments are still being run to verify correctness, but please take a look and review. Closes #254


dapatil211:
Results for runs on various mujoco environments with DDPG and TD3:
https://wandb.ai/chandar-rl/Hive/reports/DDPG-TD3-Results--VmlldzoxOTUzMTgx?accessToken=qsulda8jmwqcgtu32hz12m3thi1tdmwcuqizg9tgb2h4fgxwdz3f5mgzxfmlxhyh

They mostly match the cleanRL [benchmark](https://wandb.ai/cleanrl/cleanrl.benchmark/reports/Mujoco--VmlldzoxODE0NjE), our runs are slightly better in some cases.

hive/agents/td3.py
```diff
@@ -0,0 +1,419 @@
...
+        super().__init__(observation_space, action_space, id)
+        self._device = torch.device("cpu" if not torch.cuda.is_available() else device)
+        self._state_size = (
+            stack_size * self._observation_space.shape[0],
+            *self._observation_space.shape[1:],
+        )
+        self._action_min = self._action_space.low
+        self._action_max = self._action_space.high
+        self._action_scaling = 0.5 * (self._action_max - self._action_min)
+        self._scale_actions = np.isfinite(self._action_scaling).all()
```
hnekoeiq, line 130: Handle `tanh` when self._action_max or self._action_min is not defined.

hive/agents/qnets/td3_heads.py
```diff
@@ -0,0 +1,109 @@
...
+                for critic in critics
+            ]
+        )
+
+    def forward(self, obs, actions):
+        obs = self.network(obs)
+        obs = torch.flatten(obs, start_dim=1)
+        actions = torch.flatten(actions, start_dim=1)
+        x = torch.cat([obs, actions], dim=1)
+        if self._n_critics == 1:
```
hnekoeiq: Do you need this condition?

hive/agents/td3.py
```diff
@@ -0,0 +1,419 @@
...
+        self._critic_optimizer = critic_optimizer_fn(self._critic.parameters())
+        self._actor_optimizer = actor_optimizer_fn(self._actor.parameters())
+        if replay_buffer is None:
+            replay_buffer = CircularReplayBuffer
+        self._replay_buffer = replay_buffer(
+            observation_shape=self._observation_space.shape,
+            observation_dtype=self._observation_space.dtype,
+            action_shape=self._action_space.shape,
+            action_dtype=self._action_space.dtype,
+            gamma=discount_rate,
```
hnekoeiq, line 147: Just a reminder that this should be done for other agents too.

hive/agents/td3.py
```diff
@@ -0,0 +1,419 @@
...
+        # Calculate action and add noise if training.
+        observation = torch.tensor(
+            np.expand_dims(observation, axis=0), device=self._device
+        ).float()
+        action = self._actor(observation)
+        if self._training:
+            noise = torch.randn_like(action, requires_grad=False) * self._action_noise
+            action = action + noise
+        action = action.cpu().detach().numpy()
+        action = self.unscale_actions(np.expand_dims(action, axis=0))
```
hnekoeiq: Conditions have to be checked.

hive/agents/ddpg.py
```diff
@@ -0,0 +1,124 @@
...
+    """
+    An agent implementing the DDPG algorithm. It is implemented by fixing the
+    n_critics, policy_update_frequency, target_noise, and target_noise_clip parameters
+    of the :py:class:`~hive.agents.td3.TD3` agent.
+    """
+
+    def __init__(
+        self,
+        observation_space,
+        action_space,
```
alirahkay: Why do these not have type annotations?

hive/agents/qnets/td3_heads.py
```diff
@@ -0,0 +1,109 @@
...
+    :obj:`representation_network` and :obj:`actor_net`, and adds a final
+    :py:class:`~torch.nn.Linear` layer to compute the action."""
+
+    def __init__(
+        self,
+        representation_network: torch.nn.Module,
+        actor_net: FunctionApproximator,
+        network_output_shape,
+        action_shape,
+        use_tanh=True,
```
alirahkay, line 20: You can just pass in the activation function (based on recent changes to dev). So this way gives more flexibility, and it's much more consistent with the current state of the codebase.

Something like, if the activation function is not None, then add that nonlinearity to the end of `actor_modules`.

hive/agents/qnets/td3_heads.py
```diff
@@ -0,0 +1,109 @@
...
+        x = self.actor(x)
+        return torch.reshape(x, (x.size(0), *self._action_shape))
+
+
+class TD3CriticNetwork(torch.nn.Module):
+    def __init__(
+        self,
+        representation_network: torch.nn.Module,
+        critic_net: FunctionApproximator,
+        network_output_shape,
```
alirahkay: Type annotations?

alirahkay (changes requested):
Overall it looks good to me. 
I added my comment about the `tanh`. 

Also, as a final matter, did you run new experiments using the right number of testing episodes and add them to your WandB dashboard?

hive/agents/qnets/td3_heads.py
dapatil211: We kind of need it to be tanh because the action scaling relies on it being between -1 and 1

dapatil211:
@alirahkay Runs are added to the baseline report with the correct number of test episodes. Added type annotations.
@hnekoeiq @sriyash421  if one of you could also review.

alirahkay (approved):
LGTM!

hive/agents/td3.py
```diff
@@ -0,0 +1,424 @@
...
+            network = torch.nn.Identity()
+        else:
+            network = representation_net(self._state_size)
+        network_output_shape = calculate_output_dim(network, self._state_size)
+        self._actor = TD3ActorNetwork(
+            network,
+            actor_net,
+            network_output_shape,
+            self._action_space.shape,
+            self._scale_actions,
```
sriyash421, line 195: I guess self._scale_actions specifies if we need to scale it i.e. use tanh right? Different variable names is causing some confusion.

hive/agents/td3.py
```diff
@@ -0,0 +1,424 @@
...
+            update_info: Contains the information from the current timestep that the
+                agent should use to update itself.
+        """
+        if self._reward_clip is not None:
+            update_info["reward"] = np.clip(
+                update_info["reward"], -self._reward_clip, self._reward_clip
+            )
+        preprocessed_update_info = {
+            "observation": update_info["observation"],
+            "action": self.scale_action(update_info["action"]),
```
sriyash421, line 254: So the actor gives us an action between [-1,1] using tanh, but we unscale it to fit the action_space range, then for updating the buffer we scale down using the self.scale function. Two different scaling methods are used for the same thing. @dapatil211 this doesnt seem to be fully correct. My main concern is that do we return to exactly the same action values after the tanh, unscaling and scaling process.

hive/agents/td3.py
```diff
@@ -0,0 +1,424 @@
...
+                        self._actor.parameters(), self._grad_clip
+                    )
+                self._actor_optimizer.step()
+                self._update_target()
+                if self._logger.should_log(self._timescale):
+                    self._logger.log_scalar("actor_loss", actor_loss, self._timescale)
+
+    def _update_target(self):
+        """Update the target network."""
+        for (network, target_network) in [
```
sriyash421, line 382: Should we add flags for hard/soft updates? Some works actually use them in their hyperparameter search

hive/agents/td3.py
```diff
@@ -0,0 +1,424 @@
...
+            observation: The current observation.
+        """
+
+        # Calculate action and add noise if training.
+        observation = torch.tensor(
+            np.expand_dims(observation, axis=0), device=self._device
+        ).float()
+        action = self._actor(observation)
+        if self._training:
+            noise = torch.randn_like(action, requires_grad=False) * self._action_noise
```
sriyash421, line 294: No changes but a couple of comments.
1. File structure and naming has to be changed. I guess that can be handled  in a later PR.
2. Feature request here! Add an exploration class that can handle different exploration techniques? Uniform, gaussian, boltzmann

Rest everything looks good!

sriyash421 (approved):

PR #276, opened 2022-05-04
Add num_updates_per_train_step hyperparameter

alirahkay:
Previously we did not have this hyperparameter. It's needed in the Atari100K benchmark for the OverTrainedRainbow. The reason for adding to the general algorithms is that I think it's kind of an important hyperparameter in RL algorithms in general.  I'll be glad to hear your comments on this.

### DIFF
--- a/hive/agents/dqn.py
+++ b/hive/agents/dqn.py
@@ -119,7 +119,7 @@ class DQNAgent(Agent):
         if replay_buffer is None:
             replay_buffer = CircularReplayBuffer
         self._replay_buffer = replay_buffer(observation_shape=self._obs_dim)
-        self._discount_rate = discount_rate ** n_step
+        self._discount_rate = discount_rate**n_step
         self._grad_clip = grad_clip
         self._reward_clip = reward_clip
         self._target_net_soft_update = target_net_soft_update
--- a/hive/agents/dqn.py
+++ b/hive/agents/dqn.py
@@ -119,7 +119,7 @@ class DQNAgent(Agent):
         if replay_buffer is None:
             replay_buffer = CircularReplayBuffer
         self._replay_buffer = replay_buffer(observation_shape=self._obs_dim)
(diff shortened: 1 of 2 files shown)


hive/agents/dqn.py
```diff
@@ -98,6 +99,8 @@ def __init__(
                 to be used during testing phase.
             min_replay_history (int): How many observations to fill the replay buffer
                 with before starting to learn.
+            num_updates_per_train_step (int): indicates the number of consecutive times the agent needs to
```
dapatil211, line 102: Fix line length. Should be less than 88

hive/agents/rainbow.py
```diff
@@ -103,6 +104,8 @@ def __init__(
                 to be used during testing phase.
             min_replay_history (int): How many observations to fill the replay buffer
                 with before starting to learn.
+            num_updates_per_train_step (int): indicates the number of consecutive times the agent needs to
```
dapatil211, line 107: fix line length

hive/agents/rainbow.py
```diff
@@ -272,66 +276,72 @@ def update(self, update_info):
...
+                    self._replay_buffer.update_priorities(batch["indices"], td_errors)
+                    loss *= batch["weights"]
+                loss = loss.mean()
+
+                if self._logger.should_log(self._timescale):
+                    self._logger.log_scalar(
+                        "train_loss",
+                        loss,
+                        self._timescale,
+                    )
```
dapatil211, line 338: This will cause issues. logger.should_log does not update the step. So a bunch of training losses will be logged at the same step. Solution is to either aggregate them (as numpy) and log them outside, or update the step in the internal loop. I think I prefer the aggregation solution. Same issue for DQN.

dapatil211 (changes requested):
You also need to add the parameter to legal_moves_rainbow.

alirahkay:
I'd close this for now; we can pick it up later.

PR #280, opened 2022-05-18
activation functions added to registry and mlps

sriyash421:
Pytorch activation functions added to the registry. Now, can be used as arguments when defining an MLP.

A sanity run on DQN using Tanh as the activation: https://wandb.ai/sriyash421/Hive/runs/d3jcxb3c?workspace=user-sriyash421

The config can be modified as this:
```
representation_net:
      name: 'MLPNetwork'
      kwargs:
        hidden_units: [256, 256]
        activation_fn:
          name: 'Tanh'

```

### DIFF
--- a/hive/agents/qnets/mlp.py
+++ b/hive/agents/qnets/mlp.py
@@ -1,11 +1,11 @@
 from functools import partial
 from typing import List, Tuple, Union
-from hive.utils.utils import ActivationFn
 
 import numpy as np
 import torch
 from torch import nn
 
+from hive.utils.utils import ActivationFn
 from hive.agents.qnets.noisy_linear import NoisyLinear
 
 
--- a/hive/agents/qnets/mlp.py
+++ b/hive/agents/qnets/mlp.py
@@ -1,5 +1,6 @@
 from functools import partial
 from typing import List, Tuple, Union
+from hive.utils.utils import ActivationFn
 
 import numpy as np
 import torch
@@ -20,6 +21,7 @@ class MLPNetwork(nn.Module):
         self,
         in_dim: Tuple[int],
         hidden_units: Union[int, List[int]] = 256,
+        activation_fn: ActivationFn = None,
         noisy: bool = False,
         std_init: float = 0.5,
     ):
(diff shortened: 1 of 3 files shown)


alirahkay (approved):
LGTM! 

Thanks for this PR, by the way.

hive/agents/qnets/mlp.py
```diff
@@ -1,5 +1,6 @@
 from functools import partial
 from typing import List, Tuple, Union
+from hive.utils.utils import ActivationFn
```
dapatil211: Fix import order

dapatil211 (changes requested):

hive/agents/qnets/mlp.py
```diff
@@ -5,6 +5,7 @@
 import torch
 from torch import nn
 
+from hive.utils.utils import ActivationFn
```
dapatil211: Sorry to ask for this again, but order should be:
```
from functools import partial
from typing import List, Tuple, Union

import numpy as np
import torch
from torch import nn

from hive.agents.qnets.noisy_linear import NoisyLinear
from hive.utils.utils import ActivationFn
```

It's minor, but might as well get it correct.

dapatil211 (changes requested):

hive/agents/qnets/mlp.py
sriyash421: Will do that. Just so that I don't do it again, is there a convention to follow? Considering length of import statement or something else?

dapatil211: its just what isort does. It's sections off the imports into (1) native python imports (2) external library imports (3) internal library imports. Each of these sections is sorted alphabetically. You can configure it with vscode, and there's a command in vscode to run organize imports to do it for you (sometimes vscode doesn't get the path of the higher level module correct so it considers section 3 to be part of section 2, if that happens, you might have to configure vscode isort)

dapatil211 (approved):

hive/agents/qnets/mlp.py
sriyash421: Got it. Thanks!!

PR #281, opened 2022-05-19
Fix bug regarding the number of test episodes in run_testing

alirahkay:
Previously, the actual number of testing episodes would have been `self._test_episodes+1`. This PR aims to fix this issue.

### DIFF
--- a/hive/runners/base.py
+++ b/hive/runners/base.py
@@ -156,7 +156,7 @@ class Runner(ABC):
         self.train_mode(False)
         aggregated_episode_metrics = self.create_episode_metrics().get_flat_dict()
         episodes = 0
-        while episodes <= self._test_episodes:
+        while episodes < self._test_episodes:
             episode_metrics = self.run_episode()
             episodes += 1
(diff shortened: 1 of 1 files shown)


hive/runners/base.py
```diff
@@ -156,7 +156,7 @@ def run_testing(self):
         self.train_mode(False)
         aggregated_episode_metrics = self.create_episode_metrics().get_flat_dict()
         episodes = 0
-        while episodes <= self._test_episodes:
+        while episodes < self._test_episodes:
```
dapatil211: Could you actually just make this a for loop? I think this bug happened because we changed to test steps and then tried reverting back to test episodes but didn't do the reversion properly

dapatil211 (changes requested):

hive/runners/base.py
alirahkay: done

dapatil211 (approved):

sriyash421 (approved):
LGTM. Thanks!

PR #282, opened 2022-05-19
Revert "activation functions added to registry and mlps"

sriyash421:
Reverts chandar-lab/RLHive#280


sriyash421:
reverting this PR, as it uses  one feature from PyTorch >=1.9 and a lot of ongoing projects are using 1.8 or 1.8.1, so it would disrupt them. Will create a new pr removing that feature.

alirahkay (approved):

PR #286, opened 2022-05-24
Atari100k, DER, and OTRainbow

alirahkay:


alirahkay:
I just have not yet gotten the results for the OTRainbow. 
Please **review** the PR, but don't merge it until I make the results ready.

hive/configs/atari100k/OTRainbow.yml
```diff
@@ -0,0 +1,95 @@
...
+        on_value: True
+        period: 500
+    epsilon_schedule:
+      name: 'LinearSchedule'
+      kwargs:
+        init_value: 1.0
+        end_value: .01
+        steps: 50000
+    test_epsilon: .001
+    min_replay_history: 20000
```
TongTongX, line 71: I wonder if this is right. Do we want to start learning after one-fifth of the training phase?

alirahkay:
Based on my experiments (26 games all 5 seeds), we have:

**DER**
DER baseline, mean: 0.28543676671224993, median: 0.16136157795467734
DER RLHive, mean: 0.29438721585036576, median: 0.1904422041798461

**OTRainbow**
OTRainbow baseline, mean: 0.26407827599681083, median: 0.2036846538468431
OTRainbow RLHive, mean: 0.2385813492674032, median: 0.1378254499932332

hive/configs/atari100k/OTRainbow.yml
alirahkay: Great catch.

As we talked last time, I double-checked the config that was used for the dopamine's OTRainbow and this is the same number that is used there. 
https://github.com/google/dopamine/blob/master/dopamine/labs/atari_100k/configs/OTRainbow.gin
(Also, it is the same number in the respective paper: https://arxiv.org/pdf/2003.10181.pdf)

TongTongX: Thanks for clarifying.

dapatil211:
I think this may have been a problem with how terminal states were being handled before. Should be fixed now? Could you try rerunning with latest code?

PR #289, opened 2022-06-15
Update random.py

dapatil211:


MHashemzadeh (commented):
The code works now but regarding the training part, is that supposed that the training flag should be Flase by the user in the config file? Ideally, it is more practical to use a random agent that does not train the network by default.

hnekoeiq (approved):
LGTM assuming tests will be passed.

MHashemzadeh (approved):

PR #291, opened 2022-07-19
Update td3.py

dapatil211:
Only clamp actions to [-1,1] if scaling is active.

### DIFF
--- a/hive/agents/td3.py
+++ b/hive/agents/td3.py
@@ -332,9 +332,9 @@ class TD3(Agent):
                 noise = torch.clamp(
                     noise, -self._target_noise_clip, self._target_noise_clip
                 )
-                next_actions = torch.clamp(
-                    self._target_actor(*next_state_inputs) + noise, -1, 1
-                )
+                next_actions = self._target_actor(*next_state_inputs) + noise
+                if self._scale_actions:
+                    next_actions = torch.clamp(next_actions, -1, 1)
                 next_q_vals = torch.cat(
                     self._target_critic(*next_state_inputs, next_actions), dim=1
                 )
--- a/hive/agents/td3.py
+++ b/hive/agents/td3.py
@@ -127,6 +127,8 @@ class TD3(Agent):
         self._action_max = self._action_space.high
         self._action_scaling = 0.5 * (self._action_max - self._action_min)
         self._scale_actions = np.isfinite(self._action_scaling).all()
+        self._action_min_tensor = torch.as_tensor(self._action_min, device=self._device)
+        self._action_max_tensor = torch.as_tensor(self._action_max, device=self._device)
         self._init_fn = create_init_weights_fn(init_fn)
         self._n_critics = n_critics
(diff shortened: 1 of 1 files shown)


hive/agents/td3.py
```diff
@@ -332,9 +332,9 @@ def update(self, update_info):
                 noise = torch.clamp(
                     noise, -self._target_noise_clip, self._target_noise_clip
                 )
-                next_actions = torch.clamp(
-                    self._target_actor(*next_state_inputs) + noise, -1, 1
-                )
+                next_actions = self._target_actor(*next_state_inputs) + noise
```
sriyash421, line 336: should the actions not be scaled/clamped to [action_min, action_max] if the scaling is inactive?

dapatil211: With this code, scaling is only inactive if action_min/action_max is infinite. The only time we might want to clamp where we don't is when only one of action_min or action_max is infinite and the other is finite. In that case, we don't do any scaling or clamping, but I think that scenario is unlikely and addressing would just make the code a little unclear. Do you think it happens enough that we should address it?

alirahkay (approved):
LGTM!

hive/agents/td3.py
sriyash421: arent there envs where the action range might be [0, \infinity)?

hive/agents/td3.py
```diff
@@ -127,6 +127,8 @@ def __init__(
         self._action_max = self._action_space.high
         self._action_scaling = 0.5 * (self._action_max - self._action_min)
         self._scale_actions = np.isfinite(self._action_scaling).all()
+        self._action_min_tensor = torch.as_tensor(self._action_min, device=self._device)
```
sriyash421, line 130: @dapatil211  I think you can just keep self._action_min_tensor and maybe change all numpy operations that use the self._action_min to torch operations, and then change the action tensor to numpy using .numpy() [as opposed to first converting tensor to numpy and then using numpy operations]

I believe this will improve readability

dapatil211: Actually I just realized why I didn't do this. When I put the scale the action before putting it into the replay buffer, it's a numpy array. So I'd have to convert it to a torch tensor (on device) and then back to a numpy array.

sriyash421: I get the issue, so it seems we need to have a different variable.

Another point of improvement is that in the framework actions which are generated from the agent are passed to the runner and then are received again in the dict. Maybe we should avoid this redundancy?

sriyash421 (approved):

hive/agents/td3.py
sriyash421: I get the issue, so it seems we need to have a different variable.

Another point of improvement is that in the framework actions which are generated from the agent are passed to the runner and then are received again in the dict. Maybe we should avoid this redundancy?

dapatil211: I think for now it's ok to have that. Especially because to implement the parallel env stuff, i am planning on making the agents stateless.

PR #292, opened 2022-07-20
Update td3.py

dapatil211:
The critic loss function wasn't being used

### DIFF
--- a/hive/agents/td3.py
+++ b/hive/agents/td3.py
@@ -4,7 +4,6 @@ import os
 import gym
 import numpy as np
 import torch
-from torch.nn.functional import mse_loss
 
 from hive.agents.agent import Agent
 from hive.agents.qnets.base import FunctionApproximator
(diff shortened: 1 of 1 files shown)


sriyash421 (dismissed):

sriyash421 (commented):
Given that distributional networks are gaining some traction now, should we test the implementation with distributional value functions?

alirahkay (approved):
LGTM.

dapatil211:
@sriyash421 I am confused, is your comment with respect to the current PR?

sriyash421:
@dapatil211 yes. I think I didnt make it clear. My main doubt was that can the td3 agent directly support distributional RL if I make changes  in the config; network and loss_fn mainly?

dapatil211:
Unlikely, a distributional agent needs to be able to access and separately handle the distribution over the action space as well as the actions themselves. See the changes made to the rainbow agent. Either way, it's out of scope for this PR.

sriyash421 (approved):

PR #299, opened 2022-09-29
updated the gym version to v26

Prishruit:


hive/envs/atari/atari.py
```diff
@@ -90,8 +90,9 @@ def step(self, action=None):
         info = {}
 
         for time_step in range(self.frame_skip):
-            _, reward, done, info = self._env.step(action)
+            _, reward, term, trunc, info = self._env.step(action)
```
dapatil211: use full names for the variables

hive/envs/gym_env.py
```diff
@@ -21,6 +21,7 @@ def __init__(self, env_name, num_players=1, **kwargs):
         """
         self.create_env(env_name, **kwargs)
         super().__init__(self.create_env_spec(env_name, **kwargs), num_players)
+        self._seed = seeder.get_new_seed("environment")
```
dapatil211: This line shouldn't be here. This file should not be calling the seeder. The seed should be stored in the seed() method

hive/runners/base.py
```diff
@@ -42,7 +42,6 @@ def __init__(
             test_episodes (int): How many episodes to run testing for.
         """
         self._environment = environment
-        self._environment.seed(seeder.get_new_seed("environment"))
```
dapatil211: This line should not be deleted.

dapatil211 (changes requested):

dapatil211 (changes requested):
You also need to update the requirements.txt file and the setup.py file with the appropriate version of gym. This will also fix the unit tests ideally.

hive/runners/base.py
Prishruit: resolved.

hive/envs/gym_env.py
Prishruit: resolved.

hive/envs/atari/atari.py
Prishruit: resolved.

tests/hive/agents/test_dqn.py
```diff
@@ -259,6 +258,7 @@ def test_create_agent_with_configs(env_spec):
 
 
 def test_train_step(agent_with_mock_optimizer):
+    # ipdb.set_trace()
```
dapatil211: delete

hive/runners/single_agent_loop.py
```diff
@@ -85,19 +93,69 @@ def run_one_step(self, observation, episode_metrics):
...
+        while not (terminated or truncated) and steps < self._max_steps_per_episode:
+
+            if steps == self._max_steps_per_episode - 1:
+                terminated, truncated, observation = self.run_end_step(
+                    observation, episode_metrics
+                )
+            else:
+                terminated, truncated, observation = self.run_one_step(
+                    observation, episode_metrics
+                )
```
dapatil211: Could you write this logic in a different way? Something like:
```
while not (terminated or truncated) and steps < self._max_steps_per_episode - 1:
    terminated, truncated, observation = self.run_one_step(
        observation, episode_metrics
    )
self.run_end_step(observation, episode_metrics)
```

hive/runners/single_agent_loop.py
```diff
@@ -85,19 +93,69 @@ def run_one_step(self, observation, episode_metrics):
...
+        agent = self._agents[0]
+        stacked_observation = self._transition_info.get_stacked_state(
+            agent, observation
+        )
+
+        action = agent.act(stacked_observation)
+        next_observation, reward, terminated, _, _, other_info = self._environment.step(
+            action
+        )
+        truncated = True
```
dapatil211: I think this should actually be `truncated = not terminated`

hive/runners/single_agent_loop.py
```diff
@@ -85,19 +93,69 @@ def run_one_step(self, observation, episode_metrics):
...
+        return terminated, truncated, next_observation
+
+    def run_end_step(self, observation, episode_metrics):
+        """Run the final step of an episode.
+
+        After an episode ends, set the truncated value to true.
+
+        Args:
+            terminated (bool): Whether this episode ended.
+            truncated (bool): Whether this episode was truncated.
```
dapatil211: These are not arguments for this method.

hive/runners/multi_agent_loop.py
```diff
@@ -104,22 +109,22 @@ def run_one_step(self, observation, turn, episode_metrics):
...
         Args:
             episode_metrics (Metrics): Keeps track of metrics for current episode.
-            done (bool): Whether this step was terminal.
+            truncated (bool): Whether this step was terminal.
 
         """
         for agent in self._agents:
             if self._transition_info.is_started(agent):
-                info = self._transition_info.get_info(agent, done=done)
+                info = self._transition_info.get_info(agent, truncated=truncated)
```
dapatil211: You need to send both truncated and terminated. Also this shouldn't run? I don't think you updated TransitionInfo with the new arguments.

hive/runners/multi_agent_loop.py
```diff
@@ -130,19 +135,23 @@ def run_end_step(self, episode_metrics, done=True):
...
         # Run the loop until the episode ends or times out
-        while not done and steps < self._max_steps_per_episode:
-            done, observation, turn = self.run_one_step(
+        while not (terminated or truncated) and steps < self._max_steps_per_episode:
+            terminated, truncated, observation, turn = self.run_one_step(
                 observation, turn, episode_metrics
             )
             steps += 1
+            if steps == self._max_steps_per_episode:
+                truncated = True
```
dapatil211: again, I think `truncated=not terminated` here

dapatil211 (changes requested):

hive/runners/single_agent_loop.py
Prishruit: correct if I am wrong, but my understanding is that run_end_step should only be called if we have reached the penultimate step and the environment has not terminated or truncated, so I am planning to do the following which is equivalent to what I wrote before.

```
while not (terminated or truncated) and steps < self._max_steps_per_episode-1:
            terminated, truncated, observation = self.run_one_step(
                    observation, episode_metrics
                )
            steps += 1

        if not (terminated or truncated):
            self.run_end_step(observation, episode_metrics)
```

hive/runners/multi_agent_loop.py
Prishruit: done.

hive/runners/multi_agent_loop.py
Prishruit: done.

hive/runners/single_agent_loop.py
Prishruit: done.

hive/runners/single_agent_loop.py
Prishruit: done.

tests/hive/agents/test_dqn.py
Prishruit: done.

hive/runners/multi_agent_loop.py
```diff
@@ -104,22 +109,23 @@ def run_one_step(self, observation, turn, episode_metrics):
...
             episode_metrics (Metrics): Keeps track of metrics for current episode.
-            done (bool): Whether this step was terminal.
+            terminated (bool): Whether this step was terminal.
+            truncated (bool): Whether this step was terminal.
 
         """
         for agent in self._agents:
             if self._transition_info.is_started(agent):
-                info = self._transition_info.get_info(agent, done=done)
+                info = self._transition_info.get_info(agent, done=(terminated or truncated))
```
dapatil211: You need to modify the arguments to this function from taking done to taking both terminated and truncated. Otherwise, terminated and truncated will not be in the resulting dictionary.

dapatil211 (changes requested):
1. Anytime gym.make() is called, you need to call it as:
```
env = gym.make("CartPole-v1", apply_api_compatibility=True)
```
ie, you need to pass the apply_api_compatibility parameter.
2. Can you also modify the step functions of the multiagent environments to also return the extra truncation parameter? For marlgrid, this means if you aren't calling the gym make function in create_env, you wrap it in the `StepAPICompatibility` wrapper. For pettingzoo, it should just involve updating to a newer version of pettingzoo, which should automatically return termination and truncation. You should also make sure the multiagent examples run.

hive/configs/marlgrid/dqn.yml
```diff
@@ -4,8 +4,8 @@ train_steps: 500000
 test_frequency: 1000
 test_num_episodes: 10
 self_play: False
-num_agents: &num_agents 2
-stack_size: &stack_size 2
+num_agents: &num_agents 3
```
alirahkay, line 7: 1. How is this related to the focus of the PR?
2. If this change in the config should be pushed to dev, where is the link pointing to the generated results?
3. Please provide further explanation for these changes.

hive/envs/gym_env.py
```diff
@@ -54,19 +54,20 @@ def create_env_spec(self, env_name, **kwargs):
...
         return observation, self._turn
 
     def step(self, action):
-        observation, reward, done, info = self._env.step(action)
+        observation, reward, terminated, truncated, info = self._env.step(action)
         self._turn = (self._turn + 1) % self._num_players
-        return observation, reward, done, self._turn, info
+        return observation, reward, terminated, truncated, self._turn, info
 
     def render(self, mode="rgb_array"):
```
alirahkay: [New render method in gym](https://github.com/openai/gym/blob/6a04d49722724677610e36c1f92908e72f51da0c/gym/core.py#L153) does not have the`mode` parameter. Update this accordingly, and make sure it works.

hive/envs/gym_env.py
```diff
@@ -54,19 +54,20 @@ def create_env_spec(self, env_name, **kwargs):
         )
 
     def reset(self):
-        observation = self._env.reset()
+        observation, _ = self._env.reset(seed=self._seed)
+        self._seed = None
```
alirahkay, line 60: This line is a bit weird at first glance. And I think it makes the code a bit unreadable ([without opening the reset method in gym](https://github.com/openai/gym/blob/6a04d49722724677610e36c1f92908e72f51da0c/gym/core.py#L150)). I wonder if we could do better here. @dapatil211 Do you have any take on this thing?

alirahkay (changes requested):

hive/envs/gym_env.py
dapatil211: I agree it's not the best, but I think we have to do something like this because of gym's API change. They made the change because they don't want to force environments to have to support changing seeds in the middle of an episode, but I think maintaining the API where you separate out the seeding call is worthwhile, and this is the only way to do it. Let me know if you have a better idea.

hive/configs/marlgrid/dqn.yml
Prishruit: 1. There was a bug I guess in the config before because the same environment defined in - https://github.com/chandar-lab/RLHive/blob/dev/hive/envs/marlgrid/ma_envs/__init__.py, had num_agents 3 and that task doesn't run with num_agents 2. But yeah its not specific to this PR.
2. Since it was an error in the config I have not reproduced these results.

hive/envs/gym_env.py
Prishruit: Yeah the mode needs to be removed and we decided that it will be done in a separate pr.

alirahkay: Sorry for the confusion; I may not have stated things correctly. This line is a bug and will influence others' works. Hence, it makes the code incomplete for me, and I am not going to accept it. And, also I don't think postponing bugs to the next PRs is a good strategy in general. (Maybe you want to correct me @apsarath @dapatil211)

Based on your comment on my change request, I think you knew the bug existed before, but unfortunately, I cannot find a single comment/description discussing this issue. This behavior is very problematic because the general assumption for the reviewers is that the code is bugless, and the code's author discusses any notable things.

By the way, who are you referring to when you say *We decided ...* ?

dapatil211 (approved):

alirahkay (approved):
LGTM!

PR #306, opened 2022-11-10
restore hidden states added

hnekoeiq:


hnekoeiq:
This PR adds two features to the DRQN agent:
1- Restoring hidden states from the replay buffer
2- Burn-in frames to warm up the rnn module.

hive/agents/drqn.py
```diff
@@ -164,6 +182,46 @@ def create_q_networks(self, representation_net):
...
+                        "cell_state": self._prev_cell_state,
+                    }
+                )
+
+            elif self._rnn_type == "gru":
+                preprocessed_update_info.update(
+                    {
+                        "hidden_state": self._prev_hidden_state,
+                    }
+                )
```
kshitijkg: This works too, but it might be easier to adapt this code to other memory based architecture if we instead store "memory" that can contain any kind of memory (hidden, hidden+cell, trxl memory), etc and have a function that can pack and unpack this memory? So the user has to only modify teh pack unpack function when adding a new memory based architecture?

hnekoeiq: @kshitijkg With the most recent commit, the update and act functions are almost independent of the type of memory (I'm still calling it hidden_state because it is a DRQN agent, but we can replace everything with memory). Could you please review it again?

hive/agents/drqn.py
```diff
@@ -277,7 +316,20 @@ def update(self, update_info, agent_traj_state=None):
...
+                interm_loss = self._loss_fn(pred_qvals, q_targets)
+                mask = torch.zeros(
+                    self._replay_buffer._max_seq_len,
+                    device=self._device,
+                    dtype=torch.float,
+                )
+                mask[self._burn_frames :] = 1.0
+                mask = mask.view(1, -1)
+                interm_loss *= mask
+                loss = interm_loss.mean()
```
mrsamsami, line 380: Isn't the correct way of doing it `interm_loss.sum() / mask.sum()`?

hive/agents/drqn.py
```diff
@@ -293,3 +345,65 @@ def update(self, update_info, agent_traj_state=None):
...
+                f"received {self._rnn_type}."
+            )
+
+        return hidden_state
+
+    def get_hidden_state(self, batch):
+        if self._store_hidden == True:
+            hidden_state = (
+                torch.tensor(
+                    batch["hidden_state"][:, 0].squeeze(1).squeeze(1).unsqueeze(0),
```
mrsamsami: I'd suggest using `view()` or `reshape()` to potentially make the code cleaner.

hive/agents/drqn.py
hnekoeiq: That's true. This should be fixed after merging #339.

PR #312, opened 2022-12-06
Wandb tutorial

MHashemzadeh:


TongTongX (changes requested):
One small thing that only happens when I opened it on github: somehow the shell command cells are rendered like latex. Trying replacing `%%shell` with `%%bash`, or `%%script bash` according to [this link](https://ipython.readthedocs.io/en/stable/interactive/magics.html). Hopefully this works.
Screenshot from 2022-12-13 01-30-29

Also this branch seems to be out of sync with `dev`. Remember to merge the recent changes.

gprashant17:
Does the code assume that the config (e.g., `minigrid5_dqn.yml`) files are already defined somewhere? I am unable to run the commands because of `FileNotFoundError`.

gprashant17:
Looks fine to me

gprashant17 (approved):
Looks fine to me

MHashemzadeh:
> 



> One small thing that only happens when I opened it on github: somehow the shell command cells are rendered like latex. Trying replacing `%%shell` with `%%bash`, or `%%script bash` according to [this link](https://ipython.readthedocs.io/en/stable/interactive/magics.html). Hopefully this works. Screenshot from 2022-12-13 01-30-29
> 
> Also this branch seems to be out of sync with `dev`. Remember to merge the recent changes.

I have tried both %%bash` and `%%script bash`. It seems the font is the same as before. 


> Does the code assume that the config (e.g., `minigrid5_dqn.yml`) files are already defined somewhere? I am unable to run the commands because of `FileNotFoundError`.

Yes, you need to use your own config file. 

>

TongTongX (approved):

PR #315, opened 2023-01-11
Register runner

dapatil211:


hive/runners/base.py
```diff
@@ -108,17 +125,7 @@ def run_one_step(self, observation, turn, episode_metrics):
...
-
-        Args:
-            episode_metrics (Metrics): Keeps track of metrics for current episode.
-            done (bool): Whether this step was terminal.
-
-        """
-        return NotImplementedError
-
-    def run_episode(self):
+    def run_episode(self, environment):
```
alirahkay, line 127: It took me a while to realize that the base runner only runs `run_testing` at the end of the training. And basically, if one wants to have an evaluation at the exact step that they like, one needs to (first know this logic and second) make a call in the `run_episode` function.

I predict that someone who writes a new runner function would miss this logic. Subsequently, they'll run their code only to realize that there is only one test at the end (😅). And finally, they manage to fix their issue by looking at either Single or Multi - agent runners.


Personally, I would not mind waiting until a training episode finishes and then I'll have my evaluation. Given the large number of training steps, an evaluation in step $k$ or $k + \epsilon, (k \ggg \epsilon)$ is perfectly fine and the final results/plots not only should be the same but also unrecognizable. 

On the other hand, I get that running testing at the exact step could be important. So, I am wondering if there is a better way to do this, maybe in the `update_runner_state`. Or otherwise, I think we should be super clear in RLHive's Document on this matter.

hive/runners/multi_agent_loop.py
```diff
@@ -1,61 +1,111 @@
...
                 for.
+            seed (int): Seed used to set the global seed for libraries used by
+                Hive and seed the :py:class:`~hive.utils.utils.Seeder`.
         """
+        if seed is not None:
+            utils.seeder.set_global_seed(seed)
+        if eval_environment is None:
+            eval_environment = environment
+        environment = environment()
+        eval_environment = eval_environment() if test_frequency != -1 else None
```
alirahkay, line 65: Why is this line repeated both here and in the `single_agent_loop`? 

The logic behind the `eval_environment` is the same for all runners, is it not? It's either provided in the `__init__` or it should be the same as *train_env*.

alirahkay (changes requested):

hive/main.py
```diff
@@ -0,0 +1,34 @@
+import argparse
+from pprint import pprint
+from hive.runners.utils import load_config
+from hive.runners import get_runner
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-c", "--config")
+    parser.add_argument("-p", "--preset-config")
```
hnekoeiq: It would be nice to add some descriptions for each one of the arguments. What is `preset-config` for example?

hive/runners/base.py
```diff
@@ -41,8 +43,13 @@ def __init__(
...
             test_episodes (int): How many episodes to run testing for.
         """
-        self._environment = environment
-        self._environment.seed(seeder.get_new_seed("environment"))
+        self._train_environment = environment
+        self._train_environment.seed(seeder.get_new_seed("environment"))
+        self._eval_environment = eval_environment
+        if self._eval_environment is not None:
+            self._eval_environment.seed(seeder.get_new_seed("environment"))
+        self._environment = self._train_environment
```
hnekoeiq: Why do we have both `self._environment` and `self._train_environment`?

hive/runners/single_agent_loop.py
```diff
@@ -88,14 +117,14 @@ def run_one_step(self, observation, episode_metrics):
         if self._training:
             agent.update(copy.deepcopy(info))
 
-        self._transition_info.record_info(agent, info)
+        transition_info.record_info(agent, info)
         episode_metrics[agent.id]["reward"] += info["reward"]
         episode_metrics[agent.id]["episode_length"] += 1
         episode_metrics["full_episode_length"] += 1
 
         return terminated, truncated, next_observation
```
hnekoeiq, line 125: Maybe I didn't understand the logic completely, but shouldn't we return the `transition_info` too?

hnekoeiq (changes requested):
Some initial comments for now.

hive/runners/base.py
dapatil211: I think updating the documentation is probably the best way to do this. Different runners will likely have different requirements about when to run the testing phase, and trying to enforce it in the base class would likely mess things up.

hive/runners/multi_agent_loop.py
dapatil211: Logic is two-fold:
1. Allow more flexibility for people who want to create their custom runners that have custom behaviors relating to evaluation environments. We also want to allow the possibility that evaluation is not run and thus eval_environment is None.
2. Make the behavior of how we create these environments more visible for anyone using SingleAgentRunner

hive/runners/base.py
dapatil211: Good catch. I should remove self._environment

hive/runners/single_agent_loop.py
dapatil211: The transition_info object exists outside of this function and we are just changing the state of it, so I think this is fine, but if you can come up with a reason we should return the object, I'd be ok with changing it

hive/runners/multi_agent_loop.py
```diff
@@ -124,159 +178,40 @@ def run_end_step(self, episode_metrics, terminated=False, truncated=False):
...
         terminated, truncated = False, False
 
         # Run the loop until the episode ends or times out
-        while not (terminated or truncated) and steps < self._max_steps_per_episode:
+        while (
+            not (terminated or truncated)
+            and steps < self._max_steps_per_episode
+            and (not self._training or self._train_schedule.get_value())
+        ):
             terminated, truncated, observation, turn = self.run_one_step(
```
hnekoeiq, line 204: This comment is related to gym new update PR. I guess we are assuming that all the environments including single/multi-agent return **truncated**, right? I believe the way it's implemented now is going to break Petting Zoo envs at least since env.last() which is called in env.step() does not return truncated. Just raising this here so we add an issue maybe to fix those envs.

hnekoeiq: Since we don't have test_ma_loop yet, I just suspect that we might miss something. Could you just run a simple experiment to verify that also runs with no bug? Thanks.

hnekoeiq (approved):
Other than a general comment I added, LGTM.

hive/runners/multi_agent_loop.py
dapatil211: The new pettingzoo API does return truncated: https://pettingzoo.farama.org/content/basic_usage/. And yes I think we already ran basic multiagent experiments to make sure everything runs.

hive/runners/base.py
dapatil211: Actually I think you are right, I looked at the code again, and the reason we weren't directly calling the save or testing in the update_state method is because it would cause issues with running testing in the middle of a training episode. Since this is no longer an issue, I'll move it to update_state in the stateless_agent pr.

dapatil211: Take a look at #316 . Does that resolve this issue?

alirahkay (approved):
LGTM!

PR #316, opened 2023-01-11
Stateless agent

dapatil211:
Note, docstrings still need to be updated. Please review other things in the meantime.


hive/agents/dqn.py
```diff
@@ -264,13 +263,13 @@ def act(self, observation):
...
-            and self._state["episode_start"]
+            and state is None
         ):
             self._logger.log_scalar("train_qval", torch.max(qvals), self._timescale)
-            self._state["episode_start"] = False
-        return action
+            state = {}
+        return action, state
 
-    def update(self, update_info):
```
hnekoeiq, line 273: IMO, agent_state would be less confusing.

hive/runners/single_agent_loop.py
```diff
@@ -85,7 +85,9 @@ def __init__(
         )
         self._stack_size = stack_size
 
-    def run_one_step(self, environment, observation, episode_metrics, transition_info):
+    def run_one_step(
```
hnekoeiq, line 89: This looks weird. Do you know why Black does that?

hnekoeiq (approved):
Two minor comments. Other than that, it looks good.

hive/runners/single_agent_loop.py
dapatil211: I think it's because the line is longer now, and so it'd go above the 88 column threshold

mrsamsami:
Except for the naming, which Hadi also mentioned, everything looks good. I'd recommend changing the `state` variable name to `memory` because it basically stores information about the past and present (not only for DRQN).

mrsamsami (approved):
LGTM.

PR #326, opened 2023-02-22
Minigrid Setup

mrsamsami:


dapatil211 (changes requested):
Please also:
1. Remove the minigrid folder.
2. update the minigrid config

dapatil211:
This should be merged after #272 , since that has the support for environment wrappers. Once that's there, you should delete the MiniGridEnv class, and just register the minigrid wrappers, with names like 'minigrid.FullyObsWrapper'.

dapatil211:
@mrsamsami #272 is now merged, so can you update this PR?

hive/envs/__init__.py
```diff
@@ -1,11 +1,13 @@
...
-    from hive.envs.minigrid import MiniGridEnv
-except ImportError:
-    MiniGridEnv = None
+from minigrid import (
+    FlatObsWrapper,
+    FullyObsWrapper,
+    ImgObsWrapper,
+    RGBImgObsWrapper,
+    RGBImgPartialObsWrapper,
+)
```
dapatil211, line 10: Can you update with changes from dev branch and move these changes to gym/gym_wrappers.py, inside of a try block?
Like:
```
try:
   importing and registering minigrid wrappers
except:
   ...
```

hive/envs/__init__.py
```diff
@@ -28,7 +30,11 @@
     BaseEnv,
     {
         "GymEnv": GymEnv,
-        "MiniGridEnv": MiniGridEnv,
+        "minigrid.FullyObsWrapper": FullyObsWrapper,
+        "minigrid.FlatObsWrapper": FlatObsWrapper,
+        "minigrid.ImgObsWrapper": ImgObsWrapper,
+        "minigrid.RGBImgObsWrapper": RGBImgObsWrapper,
+        "minigrid.RGBImgPartialObsWrapper": RGBImgPartialObsWrapper,
```
dapatil211, line 37: These should be registered as gymwrappers and not environments

dapatil211 (changes requested):

PR #332, opened 2023-03-22
Padding at sampling time

mrsamsami:
`self._pad_buffer()` is removed from `RecurrentReplayBuffer`. Instead, we add zero masking to the `self.sample()`.

### DIFF
--- a/hive/replays/recurrent_replay.py
+++ b/hive/replays/recurrent_replay.py
@@ -107,14 +107,12 @@ class RecurrentReplayBuffer(CircularReplayBuffer):
         """
 
         if self._episode_start:
-            self._pad_buffer(self._max_seq_len - 1)
             self._episode_start = False
         transition = {
             "observation": observation,
             "action": action,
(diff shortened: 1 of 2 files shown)


hive/replays/recurrent_replay.py
```diff
@@ -321,4 +313,11 @@ def sample(self, batch_size):
                     num_to_access=self._max_seq_len,
                 )
 
+        mask = np.ones_like(batch["done"], dtype=bool)
+        for i in range(batch["done"].shape[0]):
+            done_indices = np.where(batch["done"][i])[0]
+            if len(done_indices) > 0:
+                mask[i, done_indices[0]:] = False  # first_done+1 ?
+        batch["mask"] = mask
+
```
dapatil211, line 322: I think you can just do something like
```
mask = np.cumsum(batch['dones'], axis=1, dtype=bool)
```

dapatil211 (changes requested):

